PR #3, opened 2022-06-29
chore: Add support for React 18

HenriqueLimas:
Note: this should NOT be a breaking change since it should also support react 16;

- Upgrade devDependencies to React 18
- Resolve react-test-render to 18 in order to fix snapshot testing ([issue](https://github.com/storybookjs/storybook/issues/17985#issuecomment-1116711479))
- Fix dialog document click event for React 18 (useEffect are sync on React causing issue on document event handlers)
- Fix props checking
- Remove testing-library/react-hook and upgrade testing-library/react to 13 (with renderHooks)
- Fix React 18 types (missing `children` on components)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,1 @@
-registry=https://registry.npmjs.org
+registry=https://registry.yarnpkg.com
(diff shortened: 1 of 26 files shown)


HenriqueLimas:
Delete in favor of #5

HenriqueLimas:
Delete in favor of https://github.com/eBay/ebayui-core-react/pull/5

PR #5, opened 2022-06-30
chore: Add support for React 18

HenriqueLimas:
Note: this should NOT be a breaking change since it should also support react 16;

- Upgrade devDependencies to React 18
- Resolve react-test-render to 18 in order to fix snapshot testing ([issue](https://github.com/storybookjs/storybook/issues/17985#issuecomment-1116711479))
- Fix dialog document click event for React 18 (useEffect are sync on React causing issue on document event handlers)
- Fix props checking
- Remove testing-library/react-hook and upgrade testing-library/react to 13 (with renderHooks)
- Fix React 18 types (missing `children` on components)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx --no-install lint-staged
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+yarn eslint-local && yarn build && yarn test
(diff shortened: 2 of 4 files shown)


src/ebay-infotip/ebay-infotip-host.tsx
```diff
@@ -8,6 +8,7 @@ type InfotipHostProps = ComponentProps<'button'> & {
     icon?: Icon;
     forwardedRef?: RefObject<HTMLAnchorElement & HTMLButtonElement>;
     variant?: Variant;
+    children?: ({ icon }) => ReactNode | ReactNode
```
darkwebdev, line 11: what does `ReactNode | ReactNode` mean?

darkwebdev (changes requested):

src/ebay-infotip/ebay-infotip-host.tsx
HenriqueLimas: Its Function `({ icon }) => React` or `ReactNode `

darkwebdev: 👍

Issue #4, opened 2022-06-30
page scrolls up to the top when the user clicks on listbox-button


montoya332:
Old default 
setTimeout(() => optionsContainerRef.current.focus(), 0)
https://user-images.githubusercontent.com/8145951/176749394-85dbf522-9fa4-4a99-9f2e-5c7ec0b90654.mov
May want add preventScroll Param
New 
setTimeout(() => optionsContainerRef.current.focus({ preventScroll: true }), 0)
https://user-images.githubusercontent.com/8145951/176749418-cf2c426d-12c1-4597-88c5-af8682ed1749.mov

darkwebdev:
fixed by #6

PR #6, opened 2022-06-30
fix(ListboxButton): page scrolls up to the top when the user clicks on listbox-button #4

montoya332:
Old default 
setTimeout(() => optionsContainerRef.current.focus(), 0)
https://user-images.githubusercontent.com/8145951/176749394-85dbf522-9fa4-4a99-9f2e-5c7ec0b90654.mov
May want add preventScroll Param
New 
setTimeout(() => optionsContainerRef.current.focus({ preventScroll: true }), 0)
https://user-images.githubusercontent.com/8145951/176749418-cf2c426d-12c1-4597-88c5-af8682ed1749.mov

### DIFF
--- a/src/ebay-listbox-button/listbox-button.tsx
+++ b/src/ebay-listbox-button/listbox-button.tsx
@@ -127,7 +127,7 @@ const ListboxButton: FC<EbayListboxButtonProps> = ({
     }
 
     const focusOptionsContainer = (focusOptions?: FocusOptions) =>
-        setTimeout(() => optionsContainerRef.current.focus(focusOptions), 0)
+        setTimeout(() => optionsContainerRef?.current?.focus(focusOptions), 0)
     const onButtonClick = () => {
         setExpanded(!expanded)
         setOptionsOpened(true)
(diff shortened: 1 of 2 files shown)


src/ebay-listbox-button/listbox-button.tsx
```diff
@@ -130,7 +130,9 @@ const ListboxButton: FC<EbayListboxButtonProps> = ({
     const onButtonClick = () => {
         setExpanded(!expanded)
         setOptionsOpened(true)
-        setTimeout(() => optionsContainerRef.current.focus(), 0)
+        setTimeout(() => optionsContainerRef.current.focus({
```
HenriqueLimas, line 133: Can you move it to a `focusOptionsContainer()` function and use it also on the [`Enter` case](https://github.com/eBay/ebayui-core-react/blob/master/src/ebay-listbox-button/listbox-button.tsx#L142)

montoya332:
Think we need to revisit setTimeout throughout component . ideally test case should look something like below to avoid white box testing but event this is causing issues
            it('focus should move to listbox', async () => {
                await new Promise((r) => setTimeout(r, 1));
                expect(component.getByRole('listbox')).toHaveFocus();
            })

HenriqueLimas:
@montoya332 we can use [waitFor]( https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) instead of the promise. Or use jest.useFakeTimers and jest.runAllPendingTimers()

HenriqueLimas (approved):

montoya332:
retest

PR #17, opened 2022-07-13
FIX Use React 16 as a dev dependency for typing issues

### LINKED ISSUE #15: Use React v16 on both dev and peer dependencies
## Problem

Currently, eBayUI Core React only support React v16 via its peer dependency:

https://github.com/eBay/ebayui-core-react/blob/e914482b5f95893b721c8f8905b07148338e95a2/package.json#L43-L47

However, on its `devDependencies` , we use v18 for `react`, `types/react`, `react-dom`, and `types/react-dom`. I think this will cause problem since eBayUI Core React will be compiled using type definitions for React v18, while the consumer is still on v16. There might be new React attributes or props on the compiled type definitions.

## Proposed Solution

One options is to use React v16 for both peer and dev dependencies, for now. When we upgrade support to React v18, then we will update both peer and dev dependencies to use v18.

HenriqueLimas:
Fixes #15

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,11 +64,12 @@
     "@storybook/storybook-deployer": "^2.8.11",
     "@testing-library/dom": "^7.29.4",
     "@testing-library/jest-dom": "^5.16.4",
-    "@testing-library/react": "^13.0.0",
-    "@testing-library/user-event": "^13.0.0",
+    "@testing-library/react": "^12",
+    "@testing-library/react-hooks": "^7",
+    "@testing-library/user-event": "^12",
     "@types/classnames": "^2.2.11",
-    "@types/react": "^18.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -100,6 +101,6 @@
     "typescript": "^3.9.5"
   },
   "resolutions": {
-    "react-test-renderer": "^18.2.0"
+    "@types/react": "^16.14.2"
```
abiyasa, line 104: Thanks, I thought we don't need it. But without this resolution, `@types/react` v17 will be installed since "@types/react-dom@<18.0.0" is depending on it.

abiyasa (approved):
Thanks, lgtm 👍

PR #57, opened 2022-07-22
BREAKING CHANGE: Rename to ui-core-react

HenriqueLimas:
- Rename package name to `@ebay/ui-core-react`
- Set version `0.1.0` so release will be published as `1.0.0`


abiyasa (approved):
Thanks for renaming the package. All good 👍

PR #61, opened 2022-07-29
Add new-window button story

darkwebdev:


src/ebay-button/__tests__/index.stories.tsx
```diff
@@ -99,3 +99,11 @@ storiesOf(`ebay-button`, module)
             <p><EbayButton priority="primary" bodyState="loading">Primary Button</EbayButton></p>
         </>
     ))
+    .add(`Open in new tab/window`, () => (
+        <EbayButton>
```
abiyasa, line 103: Why don't we add the link and the link target props? Otherwise, it's just a button with label, not even a link.

abiyasa (changes requested):
I would not recommend this pattern as it has accessibility problems.

Just repeating what Mark Lapole has mentioned on your Slack chat discussion, opening link in a new window is an anti-pattern. See eBay accessibility MIND pattern on this: https://ebay.gitbook.io/mindpatterns/antipatterns/open-new-window. As quote:

> Forcing links to open in a new window is an anti-pattern because we are taking control away from the user and forcing web-browsing behaviour upon them. If a user wishes to open a link in a new window, they can do so by using their mouse or keyboard shortcut, without our intervention

On the same document, it's stated that if we have to do it (open link in a new window), we should do it in accessible way:

### Screen reader need to warn user that the link will be open in a new window

This has to be done via offscreen text. See eBay accessibility MIND pattern on offscreen text and open link on new window: https://ebay.gitbook.io/mindpatterns/techniques/offscreen-text#link-opens-in-new-window-or-tab.

```html
<a href=”http://www.ebay.com/help” target=”_blank”>
  Help <span class=”clipped> - opens in new window or tab</span>
</a>
```

On your example, it only give warning only via icon, which is only for sighted users. When someone change the button label "Open in new tab" to "Show now", the context will be gone and it will have problems for screen-reader users. That's the reason why offscreen text is still necessary.

### Risk of introducing non-official patterns 

I don't see any eBay pattern for link with new window/tab, neither on Skin, eBay UI Core Marko, nor the Design Pattern. Maybe because this pattern is not recommended.

Your argument with cta-button is not correct since the arrow icon only indicate that the button is a link, which is not necessary open link in a new window. This pattern is existed on our Design Pattern on marketing content:


Screenshot 2022-07-30 at 14 39 52

Introducing pattern which is not officially is risky and we don't want encourage our user to do that. **So I am against to introduce this example on eBay Ui Core React.**

darkwebdev:
yep

PR #65, opened 2022-08-01
FEAT,FIX: Add destructive variant to button

### LINKED ISSUE #18: Delete button is not rendered properly with Skin v14
## Problem

Since Skin 14 and eBayUI Core v10, the delete button is not rendered properly: the button with red border is now rendered as default button with black border:

Screenshot 2022-07-13 at 20 51 27

## Causes

With Skin 14, the [delete button](https://opensource.ebay.com/skin/archive/v13.6/index.html#button) is now called destructive buttons. Instead of just 1 kind of button, [there are 3 types of destructive buttons](https://opensource.ebay.com/skin/#button):

Screenshot 2022-07-13 at 20 54 07

- The CSS modifier has changed from `btn--delete` to `btn--destructive`.
- Apart from CSS class `btn--destructive`, we need to add `btn--primary`, `btn--secondary`, or `btn--tertiary` in order to render the button correctly.

abiyasa:
This will fix #18 due to Skin v14 changes on delete/destructive button (see https://github.com/eBay/skin/pull/1673 and https://github.com/eBay/ebayui-core/pull/1628). 

Screenshot 2022-08-01 at 17 42 28

- Add new property `variant` to ebay-button. This property is optional and have 2 values (`standard` and `destructive`), the [same API as eBayUI Core Marko ebay-button](https://github.com/eBay/ebayui-core/blob/v9.1.0/src/components/ebay-button/button.stories.js#L102).
- Deprecate EbayButton priority `delete`. Since Skin v14, you have to state the button variant (`primary`, `secondary`, or `tertiary`) for destructive button.
- Update EbayButton README and storybook accordingly


abiyasa:
@HenriqueLimas  @darkwebdev  I still keep this PR as a draft since I have 2 things to clarify. 

First, I think it's better to merge https://github.com/eBay/ebayui-core-react/pull/64 first so we can have proper snapshot testing 🙂

Second, I want to deprecate EbayButton `priority="delete"` since it will be replaced by `variant="destructive"`. This is due to destructive button can now be primary, secondary, and tertiary since Skin v13.5. Before, we only have 1 type of delete button.

@HenriqueLimas @darkwebdev I can remove completely enum "delete" from EbayButton priority, but **this will cause breaking changes**. Any project that used that value will get compilation error. That's why I still keep the value "delete" from priority but it has no effect. Unfortunately, [you cannot add JSDoc to union type](https://github.com/microsoft/tsdoc/issues/164) so I had to add deprecated note on README 😞

I actually prefer to remove completely enum "delete". Easier and cleaner. Thoughts?

HenriqueLimas (approved):

PR #68, opened 2022-08-14
FIX: EbayInfotip focus host on closing infotip

### LINKED ISSUE #67: Accessibility: Infotip focus on collapse
After collapsing an infotip the focus get lost in the page but it should focus the host element

HenriqueLimas:
- Add a `hostRef` on `useTooltip` and focus on it on collapse

Fix #67

### DIFF
--- a/src/common/tooltip-utils/use-tooltip.ts
+++ b/src/common/tooltip-utils/use-tooltip.ts
@@ -1,18 +1,19 @@
-import { useState } from 'react'
+import { useState, RefObject } from 'react'
 
 type UseTooltipArgs = {
     onExpand: () => void;
     onCollapse: () => void;
     initialExpanded?: boolean;
-}
+    hostRef?: RefObject<HTMLElement>;
+};
 
 type UseTooltip = {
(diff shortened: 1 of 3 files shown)


src/common/tooltip-utils/use-tooltip.ts
```diff
@@ -26,6 +27,10 @@ export const useTooltip = ({ onExpand, onCollapse, initialExpanded = false }: Us
         if (onCollapse) {
             onCollapse()
         }
+
+        if (hostRef?.current) {
+            hostRef.current.focus()
+        }
```
abiyasa, line 33:
```suggestion
        hostRef?.current?.focus()
```

abiyasa (approved):
LGTM 👍

src/common/tooltip-utils/use-tooltip.ts
HenriqueLimas: Thanks, updated. eslint used to fail in the past, considering it as a ternary, it seems it got fixed 👍

PR #71, opened 2022-08-19
FEAT: Program badges icons

montoya332:
new type prop 
`type` | String | No | no | 'icon' or 'program-badge'  default 'icon'
Screen Shot 2022-08-19 at 10 17 57 AM


HenriqueLimas (changes requested):
We should have a separate component for program-badges called EbayProgramBadge with same or similar API as [marko version](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-program-badge)

Issue #72, opened 2022-08-22
ebay-tab: Accessibility bug due to duplicated IDs when a page has more than 1 tab


abiyasa:
## Problems

- When there are more than 1 ebay-tabs on a page, the page will get accessibility error (via Chrome lighthouse) due to duplicated ebay-tab IDs. 
- Unfortunately, there is no workaround for this issue

## How to Reproduce

- Add 2 or more EbayTabs inside a page. You can create a storybook.
- Check on Chrome LIghthouse or other accessibility checker. You will get an error like **ARIA IDs are not unique** (see https://web.dev/duplicate-id-aria/?utm_source=lighthouse&utm_medium=devtools)

## Cause

This is due to hardcoded id on nesting tab, which will always have semi-dynamic id `default-tab-{child-tab-index}`: https://github.com/eBay/ebayui-core-react/blob/26ee62ebcf4e920e78bfd33eba91c672895b398b/src/ebay-tabs/tab.tsx#L31-L42

## Possible Fix

Use prefix ID for nested tab id instead of just `default-tab-${index}`. We can get prefix id from the EbayTab id (if provided by consumer). This pattern is used on other components like dialogs or breadcrumbs.

PR #73, opened 2022-08-22
FIX: EbayTabs duplicated ids

abiyasa:
This PR will fix issue #72 (when there are more than 1 ebay-tabs on a page, the page will get accessibility error due to duplicated IDs). The solution is to support prop `id` on EbayTab which will be used to generate IDs for the nesting components (tab headers and tab panels).

Others:

- Remove disable info on EbayTabs Storybook as we don't use Storybook info addons anymore.
- Add new storybook with multiple tabs

NOTE: This is a quick fix since I did not have time to refactor the component into React hooks nor updating its storybook to use CSF format.

### DIFF
--- a/src/ebay-tabs/tab-panel.tsx
+++ b/src/ebay-tabs/tab-panel.tsx
@@ -3,6 +3,7 @@ import classNames from 'classnames'
 
 type TabPanelProps = ComponentProps<'div'> & {
     index?: number;
+    parentId?: string;
     selected?: boolean;
     fake?: boolean;
 }
@@ -10,6 +11,7 @@ type TabPanelProps = ComponentProps<'div'> & {
 const TabPanel: FC<TabPanelProps> = ({
     children,
     index,
+    parentId,
     selected,
     fake,
     className,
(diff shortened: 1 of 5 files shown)


HenriqueLimas (approved):

src/ebay-tabs/tab-panel.tsx
```diff
@@ -24,7 +26,7 @@ const TabPanel: FC<TabPanelProps> = ({
             {...rest}
             aria-labelledby={`default-tab-${index}`}
             className={classNames(className, 'tabs__panel')}
-            id={`default-tabpanel-${index}`}
+            id={`${parentId || 'default'}-tabpanel-${index}`}
```
montoya332, line 29: would prop id also work (to avoid more unique props )
            id={`${prop.id || 'default'}-tabpanel-${index}`}

abiyasa: I thought about using `prop.id` just like what you proposed, but that would diverge on how `id` prop normally works since the tab panel will output different id than the given one. I think we should keep `id` behaviour the same on every places, which will render the given id on the output HTML (see our checkbox, tooltip content, listbox button, ...). Component id might be used by `aria-labelledBy`, `aria-controls`, label's `for`, or just for `<a>` anchor.

But I see your point of not introducing a new prop `parentId` just for this purpose. We can use tab panel's id prop but we need to move the id generation logic inside tab panel into the parent component tabs. With this way, the tabs will decide what id its children will be and the tab panel will render the given id as-is.

PR #78, opened 2022-09-06
Fix tab header stealing focus from inputs

Jonaphant:
**Problem statement:** When Tabs component's props are updated, the tab header is programmatically set in focus.

**Problem use case:** Text input element within the tab body is updated by the user which results in updating the global state. The global state is then updated and propagated back down to the tabs component. Since the Tabs component sees that a prop is updated, it will take the focus from the text input and set it on the tab header.

Changes:
- Only focus on tab header when there is an update and the tab indexes have changed
- Added new storybook example with text input (similar use case as described above)

### DIFF
--- a/src/ebay-tabs/__tests__/index.stories.tsx
+++ b/src/ebay-tabs/__tests__/index.stories.tsx
@@ -176,3 +176,25 @@ storiesOf('ebay-tab', module)
             </EbayTabs>
         </div>
     </>))
+    .add('Tabs with inputs', () => {
+        const Component = () => {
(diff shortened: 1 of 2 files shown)


abiyasa (changes requested):
Unfortunately, the fix does not work since it breaks the focus indicator when user user keyboard to change the active tab.

We still need to keep the tab heading on focus when active tab changed, but should not steal the focus if the active tab is not changing.

darkwebdev:
So we need to set focus not on every update, but only when needed? I'll try to experiment with that...

PR #79, opened 2022-09-07
Convert EbayTabs to hooks

darkwebdev:
Also fixes https://github.com/eBay/ebayui-core-react/pull/78

[Storybook with example](https://opensource.ebay.com/ebayui-core-react/tabs-hooks/index.html?path=/story/ebay-tab--tabs-with-inputs)

### DIFF
--- a/src/ebay-tabs/__tests__/index.spec.tsx
+++ b/src/ebay-tabs/__tests__/index.spec.tsx
@@ -3,13 +3,13 @@ import { render, fireEvent } from '@testing-library/react'
 import { initStoryshots } from '../../../config/jest/storyshots'
 import { EbayTabs, EbayTab, EbayTabPanel } from '../index'
 
-const spy = jest.fn()
 
 describe('<EbayTabs>', () => {
     describe('on tab click', () => {
-        let wrapper
+        let wrapper, spy
 
         beforeEach(() => {
+            spy = jest.fn()
             wrapper = render(
                 <EbayTabs onTabSelect={spy}>
                     <EbayTab key="0">tab1</EbayTab>,
--- a/src/ebay-tabs/__tests__/index.spec.tsx
+++ b/src/ebay-tabs/__tests__/index.spec.tsx
@@ -3,11 +3,14 @@ import { render, fireEvent } from '@testing-library/react'
 import { initStoryshots } from '../../../config/jest/storyshots'
 import { EbayTabs, EbayTab, EbayTabPanel } from '../index'
 
+const spy = jest.fn()
+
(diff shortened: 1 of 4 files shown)


abiyasa (changes requested):
I like the unit test @Jonaphant added on his PR https://github.com/eBay/ebayui-core-react/pull/78. We should put more unit tests as part of refactoring or bug fixing.

One can test on changing selected tab:

```jsx
describe('on changing selected tab', () => {
    let wrapper

    beforeEach(() => {
        wrapper = render(
            <EbayTabs data-testid='tab-container' index={0}>
                <EbayTab key="0">tab1</EbayTab>,
                <EbayTab key="1">tab2</EbayTab>,
                <EbayTabPanel key="0">Tab panel 1</EbayTabPanel>,
                <EbayTabPanel key="1">Tab panel 2</EbayTabPanel>
            </EbayTabs>
        );
    })

    it('should show and focus on the new selected tab', () => {
        const targetTab = wrapper.container.querySelectorAll('[role="tab"]')[1]

        fireEvent.click(targetTab)

        expect(wrapper.getByText('Tab panel 1')).not.toBeVisible()
        expect(wrapper.getByText('Tab panel 2')).toBeVisible()

        expect(targetTab).toHaveFocus()
    })

    it('should change the selected tab with keyboard', () => {
        const currentTab = wrapper.container.querySelectorAll('[role="tab"]')[0]
        fireEvent.keyDown(currentTab, { key: 'ArrowRight' })

        expect(wrapper.getByText('Tab panel 2')).toBeVisible()
    })
})
```

src/ebay-tabs/__tests__/index.spec.tsx
```diff
@@ -3,23 +3,44 @@ import { render, fireEvent } from '@testing-library/react'
 import { initStoryshots } from '../../../config/jest/storyshots'
 import { EbayTabs, EbayTab, EbayTabPanel } from '../index'
 
+const spy = jest.fn()
```
abiyasa, line 6: Careful with moving the spy or mock to outside the test scope, especially for a shared spy among tests. You need to reset its status on every test, either by:

- call `spy.mockClear()` on `beforeEach()`
- or re-create the mock on `beforeEach()` (e.g `spy = jest.fn()`) since people might get confused when to call `mockClear()` vs `mockReset()`

Otherwise, if someone add another test with assertion `expect(spy).toBeCalled()`, that assertion will always pass regardless the implementation.

abiyasa (changes requested):

abiyasa (approved):

PR #80, opened 2022-09-07
FIX: Remove react-focus-lock

abiyasa:
Remove react focus lock which was removed on [a PR to fix dialog animation issue](https://github.corp.ebay.com/React/ebayui-core-react/pull/453) but accidentally [brought back on the PR to fix toast dialog accessibility issue](https://github.corp.ebay.com/React/ebayui-core-react/pull/391/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R63).

- The dialog focus trap should be handle by makeup-keyboard-trap
- This PR potentially could fix the focus issue of a dialog with a text input (the focus stays inside the input text even if user clicks on the dialog area).

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "classnames": "^2.2.6",
     "makeup-keyboard-trap": "^0.4.1",
     "makeup-screenreader-trap": "^0.4.1",
-    "react-focus-lock": "^2.9.1",
     "react-remove-scroll": "^2.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


HenriqueLimas (approved):
LGTM. Comment for others: Focus lock is already maintained on the [dialogBase](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-dialog-base/components/dialogBase.tsx#L107) by the makeup-keyboard-trap library

Issue #77, opened 2022-09-02
ebay-page-notice: Update icon size based on Skin v14.12.0


abiyasa:
Skin v14.12.0 contains changes on Page Notice which make the icon smaller and some adjusted spacing.

Since the page notice icon is rendered by our ebay-icon, we need to change our implementation to use the right icon sizes: https://github.com/eBay/ebayui-core-react/blob/e13c4e6f9beba01ce19ba8c2c94d77c3dec57357/src/ebay-page-notice/page-notice.tsx#L31-L37

Before:

![Before Skin v14.12.0](https://user-images.githubusercontent.com/1675667/184721132-dd1e218c-3275-43db-af58-a280a6ec0075.png)

After:

![After Skin v14.12.0](https://user-images.githubusercontent.com/1675667/184721771-c598f9d2-a7ba-47d9-ad03-20214ef7ec4b.png)

See related issues:

- Skin v14.12.0 release: https://github.com/eBay/skin/releases/tag/v14.2.0
- PR on Skin, updating page notice: https://github.com/eBay/skin/pull/1838

HenriqueLimas:
Available on v1.5.0

PR #84, opened 2022-09-12
MINOR: Update icon on page, section, & inline notice

abiyasa:
This will fix issue #77 , which will make the icon smaller on page notice, section notice, and inline notice, based on the Skin v14.12.0 changes.

### DIFF
--- a/src/ebay-inline-notice/inline-notice.tsx
+++ b/src/ebay-inline-notice/inline-notice.tsx
@@ -45,7 +45,7 @@ const EbayInlineNotice: FC<Props> = ({
         <div className={classNames(className, `inline-notice ${!isGeneral ? `inline-notice--${status}` : ``}`)}>
             {!isGeneral ? (
                 <span className="inline-notice__header">
-                    <EbayIcon name={`${status}-filled` as Icon} a11yText={ariaLabel} a11yVariant="label" />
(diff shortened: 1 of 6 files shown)


src/ebay-inline-notice/inline-notice.tsx
```diff
@@ -45,7 +45,7 @@ const EbayInlineNotice: FC<Props> = ({
         <div className={classNames(className, `inline-notice ${!isGeneral ? `inline-notice--${status}` : ``}`)}>
             {!isGeneral ? (
                 <span className="inline-notice__header">
-                    <EbayIcon name={`${status}-filled` as Icon} a11yText={ariaLabel} a11yVariant="label" />
+                    <EbayIcon name={`${status}FilledSmall` as Icon} a11yText={ariaLabel} a11yVariant="label" />
```
abiyasa, line 48: Unfortunately, we are still using TypeScript 3.9.10 that's why we still need to assert type to Icon.

TypeScript 4.3 got [template string type improvements](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#template-string-type-improvements), meaning, we could remove the type assertion:

```tsx
<EbayIcon name={`${status}FilledSmall`} a11yText={ariaLabel} a11yVariant="label" />
```

HenriqueLimas (approved):
Thanks for the PR

PR #85, opened 2022-09-20
FEAT: lightbox dialog without background click

montoya332:
disabling background click is not supported in Lightbox .


src/ebay-confirm-dialog/confirm-dialog.tsx
```diff
@@ -40,24 +40,28 @@ const EbayConfirmDialog: FC<Props> = ({
...
-                    aria-describedby={mainId}
-                    className="confirm-dialog__confirm"
-                >
-                    {confirmText}
-                </EbayButton>
-            </EbayDialogFooter>
+            {
+                (confirmText || rejectText) &&
+                <EbayDialogFooter>
+                    <EbayButton
```
HenriqueLimas, line 46: Should we render those button only when the text exists?

```jsx
(confirmText || rejectText) && 
<EbayDialogFooter>
  {rejectText && <EbayButton />}
  {confirmText && <EbayButton />}
</EbayDialogFooter>
```

In that way there is't a risk to render an empty button

montoya332: moved to button position logic

src/ebay-dialog-base/components/dialogBase.tsx
```diff
@@ -89,7 +89,7 @@ export const DialogBase: FC<DialogBaseProps<HTMLElement>> = ({
                 onBackgroundClick(e)
             }
         }
-        if (open) {
+        if (open && buttonPosition !== 'hidden') {
```
montoya332, line 92: https://github.com/eBay/ebayui-core/blob/6c22e9e3a9d5d546cbc7667ff4ee96a4186f42b4/src/components/components/ebay-dialog-base/component.js#L55

src/ebay-lightbox-dialog/lightbox-dialog.tsx
```diff
@@ -22,10 +22,10 @@ const EbayLightboxDialog: FC<Props> = ({
     ...rest
 }) => (
     <DialogBaseWithState
+        buttonPosition="right"
```
montoya332, line 25: moving up to override if needed

PR #102, opened 2022-10-04
Update Checkbox component to support refs

krollert-ebay:
Needed to set focus on checkbox, allows passing refs down.


montoya332 (approved):
LGTM

PR #106, opened 2022-10-04
EbaySectionTitle: Simplified html structure

### LINKED ISSUE #92: EbaySectionTitle: Update html cta structure
Update ebay-section-title html to new a.section-title__cta (remove the `<div>`)

HenriqueLimas:
- Remove CTA extra `div` and replace it with the current `a` element, following new [skin structure](https://opensource.ebay.com/skin/#section-title)
- Remove icon from CTA

Closes #92


darkwebdev:
retest

darkwebdev (changes requested):
Looks like there's also no size anymore, this doesn't make sense and needs to be removed:
image

src/ebay-section-title/cta.tsx
```diff
@@ -1,37 +1,25 @@
 import React, { ComponentProps, FC, ReactNode } from 'react'
 import classNames from 'classnames'
-import { EbayIcon, Icon } from '../ebay-icon'
 
-type Props = ComponentProps<'div'> & {
+type Props = ComponentProps<'a'> & {
   ctaText?: ReactNode;
```
montoya332, line 5: `  ctaText: ReactNode;`  
  based on the logic below

HenriqueLimas: On section-title ctaText is not required, since we are already handling the empty case, don't need to make it required

darkwebdev (approved):

PR #110, opened 2022-10-05
Page notice updates

aaron-nance:
The React version of the ebay-page-notice was missing the ability to add a dismiss button and have a CTA link with a dismiss button, which is available in the Marko version.

- Added storybook entries for adding a dismiss button and adding a dismiss button with a CTA.
- Added the new `a11yDismissText ` attribute to the documentation.
- Added unit tests around the dismiss feature, both for clicking and keyboard access. 

I also updated `testing-library/user-event` to the latest, which broke a couple tests due to breaking changes in the versions, but I updated those tests as well. 

Didn't add code to ignore the footer if a `a11yDismissText` is present, as the Marko version has.

# Screenshots

## Page Notice with Dismiss button
Screen Shot 2022-10-05 at 2 08 47 PM

## Page Notice with CTA and Dismiss button
Screen Shot 2022-10-05 at 2 09 00 PM


src/ebay-page-notice/page-notice.tsx
```diff
@@ -1,44 +1,70 @@
...
 import NoticeContent from '../common/notice-utils/notice-content'
 import { EbayNoticeContent } from '../ebay-notice-base/components/ebay-notice-content'
 import { EbayIcon, Icon } from '../ebay-icon'
 
 type Props = React.HTMLProps<HTMLElement> & {
     status?: 'general' | 'attention' | 'confirmation' | 'information',
-    'aria-label'?: string;
+    'aria-label'?: string,
+    a11yDismissText: string,
+    onDismissed?: Dispatch<boolean>
```
HenriqueLimas, line 10: I think we should use the same name as marko version for consistency: `onDismiss`
Also can we pass the event instead of the boolean to follow marko API?

src/ebay-page-notice/page-notice.tsx
```diff
@@ -1,44 +1,70 @@
...
-                        name={`${status}FilledSmall` as Icon}
-                        a11yText={ariaLabel}
-                        a11yVariant="label"
-                    />
-                </div>
-            ) : null}
-            <NoticeContent {...content.props} type="page" />
-            {children}
-        </section>
+        <>
```
HenriqueLimas, line 37: We don't need to use a fragment for this use case:

```tsx
return !dismissed && (
   <section></section>
)
```

I usually prefer to have a separate if and return null, to make it more explicity, but that is personal opinion, feel free to follow or not:

```tsx
if (dismissed) {
  return null
}

return (
   <section></section>
)
```

src/ebay-page-notice/__tests__/index.spec.tsx
```diff
@@ -45,6 +46,43 @@ describe('<EbayPageNotice>', () => {
...
+
+        beforeEach( async () => {
+            wrapper = render(
+                <EbayPageNotice status="information" aria-label="Information" a11yDismissText="Close" onDismissed={dismissMock}>
+                    <EbayNoticeContent>
+                        <p>Content</p>
+                    </EbayNoticeContent>
+                </EbayPageNotice>
+            )
+            dismissButton = await wrapper.getByRole('button', {name:'Close'});
```
HenriqueLimas, line 63: (OPTIONAL): I think we can use getByText('Close') instead, it is simpler

package.json
```diff
@@ -65,7 +65,7 @@
     "@testing-library/jest-dom": "^5.16.4",
     "@testing-library/react": "^12",
     "@testing-library/react-hooks": "^7",
-    "@testing-library/user-event": "^12",
+    "@testing-library/user-event": "^14",
```
HenriqueLimas: It would be nice if we can do this upgrade in a different PR to limit the scope of this PR.

HenriqueLimas (commented):
Thank you for the PR. Added some comments and request change (mainly to stay consistency with marko API).

Can you also add `FEAT:` in the description of the PR so we can generate a new version later

src/ebay-page-notice/page-notice.tsx
darkwebdev: I usually do
```
return !dismissed ? null : <section>...</section>
```

My point is: don't return `undefined`, components have to return either a `ReactElement` or `null`.

darkwebdev:
I've attached a storybook link at the bottom of PR description.

src/ebay-page-notice/page-notice.tsx
aaron-nance: Updated to onDismiss and now dispatches a MouseEvent.

src/ebay-page-notice/page-notice.tsx
aaron-nance: Good feedback. Updated.

HenriqueLimas (approved):

PR #119, opened 2022-10-07
Make the page notice  a11yDismissText attribute optional

aaron-nance:
The a11yDismissText attributes should have been optional, but it wasn't, which would cause issues if you were using strict typescript.


abiyasa (approved):
Thanks for the quick fix 🙂

PR #118, opened 2022-10-07
Issue 91: Update Page Notice stories to use EbayPageNoticeTitle

aaron-nance:
Many of the page-notice examples were using `strong` tags for a header instead of using `EbayPageNoticeTitle`. Updated these stories and their snapshots. 

fixes #91

PR #121, opened 2022-10-12
Issue 91: Update Page Notice stories to use EbayPageNoticeTitle #118

aaron-nance:
Many of the page-notice examples were using strong tags for a header instead of using EbayPageNoticeTitle. Updated these stories and their snapshots.

fixes https://github.com/eBay/ebayui-core-react/issues/91

This time without all the extra stuff that came along when the branch was changed to 2.0.0.


HenriqueLimas (approved):

HenriqueLimas:
Checked on storybook: https://opensource.ebay.com/ebayui-core-react/Issue91-2.0.0/index.html?path=/story/ebay-page-notice--simple-usage

PR #129, opened 2022-10-19
Adding "button" type to button

kristopherrollert:
Button without a type defaults to a "submit" action. If this drawer is hosted inside a form, it will submit the form instead of expanding. By setting the type to "button" we prevent that functionality.


kristopherrollert:
Still working out how to set this up, recreating in different PR.

PR #132, opened 2022-10-24
[#24] Fix ignored `a11yCloseText` in fullscreen dialog...

### LINKED ISSUE #24: a11yCloseText in EbayFullscreenDialog doesn't appear to see the a11y text
900ee468-084d-40d3-aeb5-b2404b8afc5c

darkwebdev:
- fix ignored a11yCloseText in fullscreen dialog
- update dialog stories with close a11y text
- make all a11y text props required for dialogs like in ebayui-marko.

fixes #24 

storybook: https://opensource.ebay.com/ebayui-core-react/fix/dialog-close-a11y/index.html?path=/story/ebay-fullscreen-dialog--always-opened

### DIFF
--- a/src/ebay-dialog-base/components/dialogBase.tsx
+++ b/src/ebay-dialog-base/components/dialogBase.tsx
@@ -35,7 +35,7 @@ export interface DialogBaseProps<T> extends HTMLProps<T> {
     top?: ReactElement;
     buttonPosition?: ButtonPosition;
     ariaLabelledby?: string;
-    a11yCloseText?: string;
+    a11yCloseText: string;
     onCloseBtnClick?: MouseEventHandler;
     onBackgroundClick?: MouseEventHandler;
     mainId?: string;
--- a/src/ebay-drawer-dialog/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/src/ebay-drawer-dialog/__tests__/__snapshots__/index.spec.tsx.snap
@@ -849,6 +849,7 @@ exports[`Storyshots ebay-drawer-dialog Opened 1`] = `
             id="dialog-title-abc123"
           />
           <button
+            aria-label="Close drawer"
             class="icon-btn drawer-dialog__close"
             type="button"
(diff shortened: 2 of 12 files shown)


src/ebay-dialog-base/components/dialogBase.tsx
```diff
@@ -58,7 +58,7 @@ export const DialogBase: FC<DialogBaseProps<HTMLElement>> = ({
     buttonPosition = 'left',
     children,
     ariaLabelledby,
-    a11yCloseText,
```
abiyasa, line 61: There was a discussion years ago why the eBay UI Core Marko try to avoid having default accessibility text. I could not find the link anymore but the problem was:

- Having the default text in English might be problematic for non-English page. For example, on German pages (`<html lang="de">`), the screenreader will read all text in German, so the English text might be read as if they're German words, which might confuse users. Not sure if this is less accessible compare to not having accessible text at all.
- To make it worse, these default english accessible text is hard to detect with accessibility checker (e.g Google Lighthouse). If the label is missing, the tool can easily point out the problems.

IMO, it's the app/consumer responsible to provide localized accessible label/text. So we should not add default text for that.

There was a discussion to make a11yText in eBay UI Core Marko mandatory, which will failed Marko template compilation, but not sure if it's implemented or not.

abiyasa: Maybe we can make it non-optional for a11yCloseText. but this is only useful if the FE app uses TypeScript. But this is a good start imo.

abiyasa (changes requested):

src/ebay-dialog-base/components/dialogBase.tsx
darkwebdev: got it, thanks Abi!

darkwebdev: fixed

abiyasa (approved):

PR #146, opened 2022-11-10
EbayMenuButton component

anch-0:
Porting over from our internal repository.
Screen Shot 2022-11-10 at 5 48 37 AM
Screen Shot 2022-11-10 at 5 48 45 AM
Screen Shot 2022-11-10 at 5 48 53 AM
Screen Shot 2022-11-10 at 5 47 58 AM
Screen Shot 2022-11-10 at 5 48 06 AM
Screen Shot 2022-11-10 at 5 48 14 AM
Screen Shot 2022-11-10 at 5 47 50 AM
Screen Shot 2022-11-10 at 5 48 29 AM
Screen Shot 2022-11-10 at 5 47 42 AM
Screen Shot 2022-11-10 at 5 48 21 AM

### DIFF
--- a/src/common/component-utils/utils.ts
+++ b/src/common/component-utils/utils.ts
@@ -10,9 +10,10 @@ export function excludeComponent(nodes: ReactNode = [], componentType: FC): Reac
     return elements.filter(({ type }) => type !== componentType)
 }
 
-export function filterByType(nodes: ReactNode = [], componentType: FC, fn?: () => void): ReactElement[] {
+export function filterByType(nodes: ReactNode = [], componentType: FC | FC[], fn?: () => void): ReactElement[] {
     const elements = Children.toArray(nodes) as ReactElement[]
-    return elements.filter(({ type }) => type === componentType)
+    const types = [componentType].flat()
+    return elements.filter(({ type }) => types.includes(type as any))
 }
 
 export function filterBy(nodes: ReactNode = [], predicate: (el: ReactElement) => boolean): ReactElement[] {
--- a/src/ebay-menu-button/README.md
+++ b/src/ebay-menu-button/README.md
@@ -1,0 +1,57 @@
+# ebay-menu-button
+
+## Demo
+[Storybook](https://pages.github.corp.ebay.com/React/ebayui-core-react/master/?path=/story/ebay-menu-button--default)
+
+## Install
+```
+yarn add ebayui-core-react
+```
+
+## Usage
+```jsx harmony
+import {
+    EbayMenuButton,
+    EbayMenuButtonItem as Item,
+    EbayMenuButtonSeparator as Separator,
+    EbayMenuButtonLabel as Label
+} from 'ebayui-core-react/ebay-menu-button';
+import '@ebay/skin/menu-button';
+<EbayMenuButton text="Menu">
+    <Item>Item 1</Item>
+    <Item>Item 2</Item>
+    <Item>Item 3</Item>
+    <Separator />
+    <Item>Log in</Item>
+</EbayMenuButton>
+// With a complext label
+<EbayMenuButton a11yText="Menu">
+    <Label><img src="menu-logo.svg" /></Label>
+    <Item>Item 1</Item>
+    <Item>Item 2</Item>
+</EbayMenuButton>
+```
+
+## EbayMenuButton Attributes
+
+Name | Type | Required | Description
+--- | --- | --- | ---
+`text` | String | No | Button label text
+`a11yText` | String | Only without a text label | A11y text for the button
+`borderless` | Boolean | No | Whether button has borders
+`fixWidth` | Boolean | No | Constrain items container width to button width
+`onExpand` | Function | No | Called when content is expanded
+`onCollapse` | Function | No | Called when content is collapsed
+`onKeydown` | Function | No | Arguments: (index: number, checked: boolean)
+`onChange` | Function | No | Arguments: (index: number, checked: boolean) for type `radio`/`checkbox`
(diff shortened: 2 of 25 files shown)


src/ebay-menu/menu.tsx
```diff
@@ -87,7 +86,7 @@ const EbayMenu: FC<Props> = ({
     }, [checkedIndexes])
 
     return (
-        <span {...rest} className={classNames(className, 'menu')}>
+        <span {...rest} className={className}>
```
abiyasa, line 89: The logic for className changed completely here. Not sure if it's intended but I think there might be a mistake to have the `menu` as default classname (see line 28)

If someone uses `<EbayMenu>` without classname, then the output will generate `<span class="menu">`, everything's fine.

However, once the user add custom className like `<EbayMenu className="menu-cancellation">`, the output will lose its `menu` class name (output: `<span class="menu-cancellation">`). 

Not sure if that's what we want since class `menu` is still required for Skin menu.

tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "module": "commonjs",
         "lib": [
             "dom",
-            "es2017",
+            "esnext",
```
abiyasa, line 7: Same as lint rule change, make sense to have a separated PR for this. There you can explain the reason behind these changes and we can focus discussing and reviewing these changes.

abiyasa (changes requested):

abiyasa:
Thanks @anch-0  for making this component 👍 I added some comments on the PR, especially in regards to `className`.

src/ebay-menu-button/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,41 @@
...
+initStoryshots({
+    config: ({ configure }) => configure(() => require('./index.stories.tsx'), module)
+})
+
+describe('<EbayMenuButton>', () => {
+    describe('on button click', () => {
+        let spy
+        beforeEach(() => {
+            spy = jest.fn()
+            spy.mockReset()
```
abiyasa, line 15: Since you always re-create the spy on beforeEach, mockReset() is not necessary here.
Anyway, good tests 👍

src/common/component-utils/utils.ts
```diff
@@ -10,9 +10,10 @@ export function excludeComponent(nodes: ReactNode = [], componentType: FC): Reac
     return elements.filter(({ type }) => type !== componentType)
 }
 
-export function filterByType(nodes: ReactNode = [], componentType: FC, fn?: () => void): ReactElement[] {
+export function filterByType(nodes: ReactNode = [], componentType: FC | FC[], fn?: () => void): ReactElement[] {
     const elements = Children.toArray(nodes) as ReactElement[]
-    return elements.filter(({ type }) => type === componentType)
+    const types = [componentType].flat()
```
abiyasa, line 15: Sorry, I just notice this later but unfortunately, array flat is still not supported on mobile Safari 11. It is available starting from Safari 12, [see browser support for array.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility). According to eBay Browser Policy, mobile Safari 11 is categorized as Tier 2 browsers but we still need support it for critical sites.

See:

- eBay Browserlist config: https://github.com/eBay/browserslist-config/blob/v2.4.1/index.js#L8
- See eBay internal wiki about browser policy

We need to manually check & unwrap `FC[]`, or use arrayflat polyfill.

abiyasa (changes requested):

src/ebay-menu-button/README.md
```diff
@@ -0,0 +1,57 @@
...
+<EbayMenuButton text="Menu">
+    <Item>Item 1</Item>
+    <Item>Item 2</Item>
+    <Item>Item 3</Item>
+    <Separator />
+    <Item>Log in</Item>
+</EbayMenuButton>
+// With a complext label
+<EbayMenuButton a11yText="Menu">
+    <Label><img src="menu-logo.svg" /></Label>
```
abiyasa, line 29: With custom label maybe?

Even it's for README, I would still suggest to use accessible example:

```tsx
<EbayMenuButton a11yText="Menu">
    <Label><img src="menu-logo.svg" alt="Plus program logo" /></Label>
```

or keep it simple like [Marko ebay-menu-button example](https://github.com/eBay/ebayui-core/tree/v10.1.0/src/components/ebay-menu-button#label-usage):

```tsx
<EbayMenuButton a11yText="Menu">
    <Label><div>Custom Label</div></Label>
```

abiyasa (changes requested):

abiyasa (changes requested):
@darkwebdev  @anch-0  I just run storybook locally. Unfortunately, the component does not work.

The menu is expanded/open by default and clicking the button will not closing the menu (menu always stay open):

Screenshot 2022-11-10 at 21 40 34

Screenshot 2022-11-10 at 21 40 39

Not sure what's wrong but it could be that the PR branch was already too long so it's outdated with the `main` branch.

- Please try to run and test the component on Storybook and compare it with the Marko version  (https://opensource.ebay.com/ebayui-core/?path=/story/buttons-ebay-menu-button--standard). 
- Apart from mouse/pointer, make sure the keyboard-only interaction and focus management works with the component. For more info on testing, see eBay OATMEAL, the accessibility testing guideline for keyboard (https://ebay.gitbook.io/oatmeal/keyboard)

darkwebdev:
@abiyasa thanks a lot! I'll take a look at the component more closely next week.

darkwebdev:
deprecating in favor of #156

PR #148, opened 2022-11-10
EbayFakeMenu component

anch-0:


darkwebdev (changes requested):
in readme:
- Please replace all github.corp links to public github links
- Please replace `ebayui-core-react` package name with `@ebay/ui-core-react`

darkwebdev (changes requested):
What's missing from the [Marko](https://opensource.ebay.com/ebayui-core/?path=/docs/building-blocks-ebay-fake-menu--standard) version:

EbayFakeMenu Props:
`type`: Can be "fake"/ "radio" / "checkbox"
`priority`: button priority, "primary" / "secondary" (default) / "none"
`itemMatchesUrl`: used in conjunction with current -- This determines whether aria-current will be page or true -- Defaults to true which gives aria-current a value of page

EbayMenuItem Props:
`current`: whether or not the href is the current href of the page

darkwebdev:
deprecating this in favor of #149

Issue #34, opened 2022-07-18
EbayFakeMenu


pcanella:
Marko version:
https://ebay.github.io/ebayui-core/?path=/story/ebay-fake-menu--basic

PR #149, opened 2022-11-11
EbayFakeMenu MVP

darkwebdev:
What's not yet supported from [Marko](https://opensource.ebay.com/ebayui-core/?path=/docs/building-blocks-ebay-fake-menu--standard) version:
EbayFakeMenu Props:
`type`: Can be "fake"/ "radio" / "checkbox"
`priority`: button priority, "primary" / "secondary" (default) / "none"

EbayMenuItem Props:
`type`: Set to "button" for fake menu-item <button>
`checked`: whether or not item is checked
`value`: the value to use with event responses for for the checked array
`badgeNumber`: used as the number to be placed in the badge
`badgeAriaLabel`: Yes (only if badge number is provided) \| passed as the aria-label directly to the badge

extracted to an issue #151 

Storybook: https://opensource.ebay.com/ebayui-core-react/EbayFakeMenu-mvp/index.html?path=/story/ebay-fake-menu--default


src/common/component-utils/utils.ts
```diff
@@ -10,9 +10,10 @@ export function excludeComponent(nodes: ReactNode = [], componentType: FC): Reac
     return elements.filter(({ type }) => type !== componentType)
 }
 
-export function filterByType(nodes: ReactNode = [], componentType: FC, fn?: () => void): ReactElement[] {
+export function filterByType(nodes: ReactNode = [], componentType: FC | FC[], fn?: () => void): ReactElement[] {
     const elements = Children.toArray(nodes) as ReactElement[]
-    return elements.filter(({ type }) => type === componentType)
+    const types = [componentType].flat()
```
anch-0, line 15: From Abi earlier ```Sorry, I just notice this later but unfortunately, array flat is still not supported on mobile Safari 11. It is available starting from Safari 12, [see browser support for array.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility). According to eBay Browser Policy, mobile Safari 11 is categorized as Tier 2 browsers but we still need support it for critical sites.

See:

eBay Browserlist config: https://github.com/eBay/browserslist-config/blob/v2.4.1/index.js#L8
See eBay internal wiki about browser policy
We need to manually check & unwrap FC[], or use arrayflat polyfill.```

darkwebdev: added polyfill

src/ebay-fake-menu/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+    describe('on menu item click', () => {
+        it('should fire onClick event', () => {
+            const onClickSpy = jest.fn()
+            const wrapper = render(
+                <EbayFakeMenu>
+                    <EbayFakeMenuItem href="#" onClick={onClickSpy}/>
+                </EbayFakeMenu>
+            )
+
+            fireEvent.click(wrapper.container.querySelectorAll('.fake-menu__item')[0])
```
abiyasa, line 20:
```suggestion
            fireEvent.click(wrapper.container.querySelectorAll('.fake-menu__item')[0])

            expect(onClickSpy).toBeCalled()            
```

You forgot the assertion

src/ebay-fake-menu/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+            const onKeyDownSpy = jest.fn()
+            const wrapper = render(
+                <EbayFakeMenu onKeyDown={onKeyDownSpy}>
+                    <EbayFakeMenuItem href="#" />
+                </EbayFakeMenu>
+            )
+
+            fireEvent.keyDown(wrapper.container.querySelectorAll('.fake-menu__item')[0])
+
+            expect(onKeyDownSpy.mock.calls[0][1]).toEqual(0)
```
abiyasa, line 34: You can simplify this with toBeCalledWith:

```suggestion
            expect(onKeyDownSpy).toBeCalledWith(expect.anything(), 0)
```

src/ebay-fake-menu/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+            const wrapper = render(
+                <EbayFakeMenu onSelect={onSelectSpy}>
+                    <EbayFakeMenuItem href="#" />
+                    <EbayFakeMenuItem href="#" />
+                </EbayFakeMenu>
+            )
+
+            fireEvent.click(wrapper.container.querySelectorAll('.fake-menu__item')[1])
+
+            expect(onSelectSpy.mock.calls[0][1]).toEqual(1)
```
abiyasa, line 49: Same here

```suggestion
            expect(onSelectSpy).toBeCalledWith(expect.anything(), 1)
```

abiyasa (approved):
Apart from minor issues on test, lgtm 👍

src/ebay-fake-menu/menu-item.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { ComponentProps, FC, MutableRefObject, useEffect } from 'react'
+import classNames from 'classnames'
+import { EbayIcon } from '../index'
+
+export type EbayFakeMenuItemProps = Omit<ComponentProps<'a'>, 'onKeyDown'> & {
+    current?: boolean;
+    disabled?: boolean;
+    itemRef?: MutableRefObject<HTMLAnchorElement>;
```
abiyasa, line 8: It's working but I never seen this pattern before. Is it better if we use boolean props like `autoFocusFirstItem` and do autofocus on first menu item if the prop is true?

- In this way, we don't need to pass a ref to the component. 
- Ref is used internally to focus the first item

Since the value from the given ref is not used by the component (in fact it's ignored and will be replaced with the ref to anchor element), it's cleaner to do it that way.

darkwebdev: you're right, replaced with `autofocus` prop

darkwebdev:
resolves #34

PR #156, opened 2022-11-17
Add ebay menu button changes

darkwebdev:


src/ebay-menu/README.md
```diff
@@ -26,7 +26,7 @@ Name | Type | Required | Description
 --- | --- | --- | ---
 `type` | String | No | Can be `radio`/`checkbox`
 `checked` | Number | No | when used with `radio` type will check the item with the corresponding index
-`onKeydown` | Function | No | arguments: (index: number, checked: boolean)
+`onKeydown` | Function | No | arguments: (index: number, checked: boolean, event)
```
abiyasa, line 29: Seems there's inconsistency here. We have param event for `onKeyDown()` but not on `onSelect()` nor `onChange()`. Same thing with eBay-fake-menu, where event was just removed from `onSelect()` in this PR.

On some components (e.g ebay-radio, ebay-select ,ebay-switch, ...), `onChange()` has param `event`. I think make sense **to always return `event` as the first parameter**, which will give more flexibility to the user. Checkout [Material UI](https://mui.com/material-ui/) or [Microsoft Fluent UI](https://developer.microsoft.com/en-us/fluentui#/controls/web) which returns event param as the first one on every handlers.

I would suggest not to remove event param on this PR, and always return `event` for ebay-menu-button. We can create another issue to change event handlers on all components, which will be breaking changes

src/ebay-fake-menu/menu.tsx
```diff
@@ -1,16 +1,12 @@
...
 import { EbayFakeMenuItemProps } from './index'
 
 type SpanProps = Omit<ComponentProps<'div'>, 'onKeyDown'>
-type Callback = (event: KeyboardEvent | MouseEvent, i: number) => void
 type Props = SpanProps & {
     itemMatchesUrl?: boolean;
-    onKeyDown?: Callback;
-    onSelect?: Callback;
+    onKeyDown?: (i: number, event: KeyboardEvent) => void
+    onSelect?: (i: number) => void;
```
abiyasa, line 9: See my previous comment to keep the event as the first param for onSelect and onKeyDown. Besides, when you change event handler signature, it will introduce breaking changes on ebay-fake-menu.

src/ebay-menu-button/menu-button.tsx
```diff
@@ -0,0 +1,172 @@
...
+    }, [expanded])
+
+    useEffect(() => {
+        setMenuId(randomId())
+    }, [])
+
+    const handleMenuKeydown = (_, __, e) => {
+        handleEscapeKeydown(e, () => {
+            setExpanded(false)
+            buttonRef.current?.focus()
```
abiyasa, line 108: Strangely, the focus does not return to the button for variant overflow.

- See storybook: https://opensource.ebay.com/ebayui-core-react/EbayMenuButton-MVP/index.html?path=/story/ebay-menu-button--variant-overflow
- Use keyboard to open the menu
- Press Esc to close the menu
- Focus does not return to the icon button (focus should be on the icon button)

abiyasa (changes requested):
Most functionality works, only focus issue on overflow type. Checkout my comment on keeping synthetic event from handlers

src/ebay-menu/README.md
darkwebdev: I put it at the very end for backwards compatibility. EbayUI marko has a discrepancy between readme and real code, so I'm not sure what's best...

darkwebdev:
retest

src/ebay-menu-button/menu-button.tsx
```diff
@@ -0,0 +1,178 @@
...
+    a11yText,
+    onClick = () => {},
+    onExpand = () => {},
+    onCollapse = () => {},
+    onChange = () => {},
+    onSelect = () => {},
+    children,
+    ...rest
+}) => {
+    const [expanded, setExpanded] = useState(defaultExpanded)
```
flakronkadriu, line 60: expanded -> isExpanded?

src/ebay-menu-button/menu-button.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+import { EbayMenuButtonItem, EbayMenuButtonLabel, EbayMenuButtonSeparator, EbayMenuButtonVariant } from '.'
+import {
+    EbayButton, EbayButtonProps,
+    EbayIcon, EbayIconButton,
+    EbayMenu, EbayMenuType, EbayMenuChangeEventHandler, EbayMenuSelectEventHandler
+} from '..'
+
+export type EbayMenuButtonProps = {
+    a11yText?: string;
```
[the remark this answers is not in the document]
darkwebdev: that's for compatibility with ebayui marko: https://opensource.ebay.com/ebayui-core/?path=/story/buttons-ebay-menu-button--standard

But yes you can use aria-label instead 👍

src/ebay-menu-button/menu-button.tsx
```diff
@@ -0,0 +1,178 @@
...
+    className?: string;
+    fixWidth?: boolean;
+    reverse?: boolean;
+    text?: string;
```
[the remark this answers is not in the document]
darkwebdev: That's ebayui marko API: https://opensource.ebay.com/ebayui-core/?path=/story/buttons-ebay-menu-button--standard

src/ebay-menu-button/menu-button.tsx
darkwebdev: We usually don't use `is` prefix, like `disabled`, `opened`, cause we always know the type of variables and props due to using typescript. In weak-typed languages adding prefix helps to understand variable's type.

src/ebay-menu-button/menu-button.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { filterByType, findComponent } from '../common/component-utils'
+import { handleEscapeKeydown } from '../common/event-utils'
+import { randomId } from '../common/random-id'
+
+import { EbayMenuButtonItem, EbayMenuButtonLabel, EbayMenuButtonSeparator, EbayMenuButtonVariant } from '.'
+import {
+    EbayButton, EbayButtonProps,
+    EbayIcon, EbayIconButton,
+    EbayMenu, EbayMenuType, EbayMenuChangeEventHandler, EbayMenuSelectEventHandler
+} from '..'
```
shpandian, line 12: we need to fix these global import right?

darkwebdev: yep, fixed!

PR #164, opened 2022-11-23
[FEAT] EbayPagination: support overflow

don-siu:
- Support `variant={overflow}`
- Added 2 dots with EbayFakeMenu
- Added Unit Tests once all the PR feedbacks are addressed 

Recording 2022-12-01 at 12 51 42


darkwebdev (changes requested):
When only second '...'-button shown then the pages in dropdown are 30-34. But when both buttons shown, then all available pages present.
image

src/ebay-button/button-expand.tsx
```diff
@@ -1,14 +1,20 @@
...
 import EbayButtonCell from './button-cell'
 import EbayButtonText from './button-text'
 import { EbayIcon } from '../ebay-icon'
 
-const EbayButtonExpand: FC = ({ children }) => (
+type Props = {
+    className?: string;
+    children: ReactNode;
+    showDropdown?: boolean;
+}
```
abiyasa, line 10: @don-siu Could you synch your feature branch with main one? I think your feature branch is out dated with the main one.

Also, you don't need to redefine children on your `Props` here since `FC` will provide type for children.

abiyasa (changes requested):
Please synch your feature branch with the remote main one. There are some outdated codes.

src/ebay-pagination/pagination.tsx
```diff
@@ -83,19 +98,46 @@ const EbayPagination: FC<PaginationProps> = ({
...
+                    firstDotItems.push(itemComponent)
+                } else if (index <= secondDotTo && index >= secondDotFrom) {
+                    secondDotItems.push(itemComponent)
+                }
+            }
 
+            if (itemType === 'page' && isDot) {
+                return (
+                    <li key={key}>
+                        <EbayFakeMenuButton className="pagination__item" text="…" borderless showDropdown={false}>
```
abiyasa, line 135: To make it more accessible, you need to change the HTML structure for the overflow menu:

- Wrap the fake menu button with a span with class `pagination__item` and add attribute `role="separator"`. Here, you added the class `pagination__item` directly on the fake menu button itself (which is not correct), thus you cannot add `role="separator"` to the fake menu button.
- Remove the text '...' from the fake menu button. The fake menu button should have `aria-label="Menu"`

Checkout the HTML structure on eBay UI Core Marko: https://opensource.ebay.com/ebayui-core/?path=/story/navigation-disclosure-ebay-pagination--interactive&args=variant:overflow or Skin https://opensource.ebay.com/skin/#pagination on **Overflow Control for Pagination**

abiyasa (changes requested):

src/ebay-pagination/pagination.tsx
don-siu: will update

src/ebay-button/button-expand.tsx
don-siu: cool @abiyasa , I will pull upstream in the coming batch of changes. 

will update for the `Props` issue

don-siu:
> When only second '...'-button shown then the pages in dropdown are 30-34. But when both buttons shown, then all available pages present. image

Good catch, will fix

src/ebay-button/button-expand.tsx
don-siu: hmm I have to define props to pass `showDropDown`. What do you mean by FC will provide type of children ?

src/ebay-pagination/pagination.tsx
don-siu: First feedback is updated'...'

I cannot remove text '...', I think `EbayFakeMenuButton` doesn't support `variant='overflow'` yet

don-siu:
will work on unit tests tomorrow

src/ebay-button/button-expand.tsx
abiyasa: @don-siu  You need to remove only the `children` from your Props. FC will provide that for you (see https://github.com/facebook/create-react-app/pull/8177).

```tsx
type Props = {
    className?: string;
    showDropdown?: boolean;
}
```

src/ebay-pagination/pagination.tsx
abiyasa: You're right, EbayFakeMenuButton did not support overflow variant. 

But now it does, the PR https://github.com/eBay/ebayui-core-react/pull/169 has been merged today. Please sync you feature branch with the main.

src/ebay-button/button-expand.tsx
```diff
@@ -1,15 +1,20 @@
...
+
+const EbayButtonExpand: FC<Props> = ({ children, showDropdown }) =>
     children ? (
         <EbayButtonCell>
             <EbayButtonText>
                 {children}
             </EbayButtonText>
-            <EbayIcon name="dropdown" />
+            {showDropdown && <EbayIcon name="dropdown" />}
```
don-siu, line 17: @abiyasa Can you explain to me how we hide the dropdown Icon here?

don-siu:
TODO: I will work on unit tests specific for 2 dots

src/ebay-button/button-expand.tsx
abiyasa: @don-siu Wait, an expand button without its expander icon is just a button, no?

I think something is not right or not necessary here. You modified ebay-button so you can display the expander button without its expander icon, only the overflow icon, right? If so, you dont have to do that anymore.

If you use the latest ebay-fake-menu-button variants overflow, it already does that for you (or check the [latest ebay-ui-core-react storybook](https://opensource.ebay.com/ebayui-core-react/main/index.html?path=/story/ebay-fake-menu-button--variants)):

Screenshot 2022-12-01 at 09 43 15

I would suggest you to remove completely `showDropdown`:

- pull the latest main branch
- Undo your changes on ebay-button
- Remove showDropdown from EbayFakeMenuButton

don-siu: updated and fixed

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ebay/ui-core-react",
-  "version": "3.7.0",
+  "version": "3.7.1",
```
msendlakowski, line 3: Minor issue, but how particular is the team about semver @abiyasa @darkwebdev ?  Technically this should be a minor version change, not a patch.

src/ebay-button/button-expand.tsx
don-siu: updated

src/ebay-pagination/pagination.tsx
don-siu: updated

package.json
don-siu: updated accordingly to Henrique's conversation

abiyasa (changes requested):
Unfortunately, there is a bug on the items on the menu. 

EDITED: this bug is not really valid since show-last variant should not show any menu. See my next comment.

When you change the selected page, especially the one closer to the overflow menu, the pages inside menu item are not correct. Please see attached video, which I generate from storybook `variant-showlast-interactive`:

https://user-images.githubusercontent.com/1072047/205364920-14056c74-2f61-4304-8f41-a64cb6dddb04.mov

I haven't check the code why this bug happens

abiyasa (changes requested):
These changes unfortunately break the pagination variant="show-last". If you run the storybook for `variant-showlast-interactive`, the drop down menu appears if you click the [...] icon.

Screenshot 2022-12-02 at 19 59 11

**Expectation**: For variant show-last, the drop down menu should never appear. See Marko pagination for variant-show-last: https://opensource.ebay.com/ebayui-core/iframe.html?args=variant:show-last&id=navigation-disclosure-ebay-pagination--interactive&viewMode=story

abiyasa (changes requested):
There is a problem on selecting page on the overflow menu.

If you click the page item from the menu, the `onSelect` is not triggered. See the following video from overflow storybook. When I select the page 13 from the menu, it reload the whole page. If I select any page item (outside the menu), the selected page is updated.

https://user-images.githubusercontent.com/1072047/205369041-f82a4f72-9ad4-4fa8-908a-76404526c670.mov

@darkwebdev  @don-siu  I think the problem is that the pagination menu is implemented using EbayFakeMenuButton, which is a menu of links, while the storybook is displaying pagination with buttons. This overflow pagination will only work on pagination with links but not for pagination with non-links.

don-siu:
> Unfortunately, there is a bug on the items on the menu.
> 
> EDITED: this bug is not really valid since show-last variant should not show any menu. See my next comment.
> 
> When you change the selected page, especially the one closer to the overflow menu, the pages inside menu item are not correct. Please see attached video, which I generate from storybook `variant-showlast-interactive`:
> 
>  pagination-overflow-bug.mov 
> I haven't check the code why this bug happens

Thanks for catching this @abiyasa , I will work this bug

don-siu:
> There is a problem on selecting page on the overflow menu.
> 
> If you click the page item from the menu, the `onSelect` is not triggered. See the following video from overflow storybook. When I select the page 13 from the menu, it reloads the whole page instead of changing the selected page to 13. If I select any page item (outside the menu), the selected page is updated (correct behaviour).
> 
>  pagination-overflow-cannot-select-item.mov 
> Checkout Marko storybook for comparison: https://opensource.ebay.com/ebayui-core/iframe.html?args=variant:overflow&id=navigation-disclosure-ebay-pagination--interactive&viewMode=story
> 
> @darkwebdev @don-siu I think the problem is that the pagination menu is implemented using EbayFakeMenuButton, which is a menu of links, while the storybook is displaying pagination with buttons. This overflow pagination will only work on pagination with links but not for pagination with non-links.

I thought there's our existing 


> There is a problem on selecting page on the overflow menu.
> 
> If you click the page item from the menu, the `onSelect` is not triggered. See the following video from overflow storybook. When I select the page 13 from the menu, it reloads the whole page instead of changing the selected page to 13. If I select any page item (outside the menu), the selected page is updated (correct behaviour).
> 
>  pagination-overflow-cannot-select-item.mov 
> Checkout Marko storybook for comparison: https://opensource.ebay.com/ebayui-core/iframe.html?args=variant:overflow&id=navigation-disclosure-ebay-pagination--interactive&viewMode=story
> 
> @darkwebdev @don-siu I think the problem is that the pagination menu is implemented using EbayFakeMenuButton, which is a menu of links, while the storybook is displaying pagination with buttons. This overflow pagination will only work on pagination with links but not for pagination with non-links.

I understand your concerns but it seems like the current pagination data structure across eBay triggers the `href` for redirection/page refresh. I am not sure if it's a common use-case for user to do a custom `event.preventDefault()` in `onSelect()` to update their local state. Please advise.

abiyasa:
> > There is a problem on selecting page on the overflow menu.
> > If you click the page item from the menu, the `onSelect` is not triggered. See the following video from overflow storybook. When I select the page 13 from the menu, it reloads the whole page instead of changing the selected page to 13. If I select any page item (outside the menu), the selected page is updated (correct behaviour).
> > pagination-overflow-cannot-select-item.mov
> > Checkout Marko storybook for comparison: https://opensource.ebay.com/ebayui-core/iframe.html?args=variant:overflow&id=navigation-disclosure-ebay-pagination--interactive&viewMode=story
> > @darkwebdev @don-siu I think the problem is that the pagination menu is implemented using EbayFakeMenuButton, which is a menu of links, while the storybook is displaying pagination with buttons. This overflow pagination will only work on pagination with links but not for pagination with non-links.
> 
> I understand your concerns but it seems like the current pagination data structure across eBay triggers the `href` for redirection/page refresh. I am not sure if it's a common use-case for user to do a custom `event.preventDefault()` in `onSelect()` to update their local state. Please advise.

@don-siu  I thought so too but why did the page refresh only happen if I click the page item inside the menu? If I click the normal page item, the page is not refreshed and the selected page item is updated.

I suspect that clicking page item inside the menu will not trigger pagination `onSelect` (it should). You have to handle `onSelect` from the EbayFakeMenuButton and trigger pagination `onSelect`. See your EbayFakeMenuButton: https://github.com/eBay/ebayui-core-react/pull/164/files#diff-3f4f4f84300864a47c6452acc3784140a5e1b3e77b005d5f0ea354ca33c73d09R130-R135

don-siu:
> Unfortunately, there is a bug on the items on the menu.
> 
> EDITED: this bug is not really valid since show-last variant should not show any menu. See my next comment.
> 
> When you change the selected page, especially the one closer to the overflow menu, the pages inside menu item are not correct. Please see attached video, which I generate from storybook `variant-showlast-interactive`:
> 
>  pagination-overflow-bug.mov 
> I haven't check the code why this bug happens

@abiyasa I have fixed the issue for `show-last` but I realized `variant=show-last` has been performed differently. Not sure if we should fix this in this or next PR. 

`ebayui-core-react`: 
https://opensource.ebay.com/ebayui-core-react/main/index.html?path=/story/ebay-pagination--variant-showlast-interactive

`ebayui-core`
https://opensource.ebay.com/ebayui-core/?path=/story/navigation-disclosure-ebay-pagination--interactive&args=variant:show-last

don-siu:
> > > There is a problem on selecting page on the overflow menu.
> > > If you click the page item from the menu, the `onSelect` is not triggered. See the following video from overflow storybook. When I select the page 13 from the menu, it reloads the whole page instead of changing the selected page to 13. If I select any page item (outside the menu), the selected page is updated (correct behaviour).
> > > pagination-overflow-cannot-select-item.mov
> > > Checkout Marko storybook for comparison: https://opensource.ebay.com/ebayui-core/iframe.html?args=variant:overflow&id=navigation-disclosure-ebay-pagination--interactive&viewMode=story
> > > @darkwebdev @don-siu I think the problem is that the pagination menu is implemented using EbayFakeMenuButton, which is a menu of links, while the storybook is displaying pagination with buttons. This overflow pagination will only work on pagination with links but not for pagination with non-links.
> > 
> > 
> > I understand your concerns but it seems like the current pagination data structure across eBay triggers the `href` for redirection/page refresh. I am not sure if it's a common use-case for user to do a custom `event.preventDefault()` in `onSelect()` to update their local state. Please advise.
> 
> @don-siu I thought so too but why did the page refresh only happen if I click the page item inside the menu? If I click the normal page item, the page is not refreshed and the selected page item is updated.
> 
> I suspect that clicking page item inside the menu will not trigger pagination `onSelect` (it should). You have to handle `onSelect` from your EbayFakeMenuButton (see line [130 on pagination.tsx](https://github.com/eBay/ebayui-core-react/pull/164/files#diff-3f4f4f84300864a47c6452acc3784140a5e1b3e77b005d5f0ea354ca33c73d09R130-R135)) and trigger pagination `onSelect` properly. Also make sense to write unit test for this.
> 
> https://github.com/eBay/ebayui-core-react/blob/4388b5c418dcf546bf85300df91d30482f3fb844/src/ebay-pagination/__tests__/index.spec.tsx#L77-L83

You're right, let me investigate more. I will probably have to update `ebayFakeMenu` to accommodate this

abiyasa:
> > Unfortunately, there is a bug on the items on the menu.
> > EDITED: this bug is not really valid since show-last variant should not show any menu. See my next comment.
> > When you change the selected page, especially the one closer to the overflow menu, the pages inside menu item are not correct. Please see attached video, which I generate from storybook `variant-showlast-interactive`:
> > pagination-overflow-bug.mov
> > I haven't check the code why this bug happens
> 
> @abiyasa I have fixed the issue for `show-last` but I realized `variant=show-last` in `ebayui-core-react` has been performed differently. Not sure if we should fix it in this or next PR.
> 
> `ebayui-core-react`: https://opensource.ebay.com/ebayui-core-react/main/index.html?path=/story/ebay-pagination--variant-showlast-interactive
> 
> `ebayui-core` https://opensource.ebay.com/ebayui-core/?path=/story/navigation-disclosure-ebay-pagination--interactive&args=variant:show-last

@don-siu  Not sure what you mean by performed differently but I agree, we should create different issue if the bug is already existed before this PR 👍 Let's focus on implementing overflow variant and make sure this new feature does not break other existing features 🙂

For an example, I just found a bug on pagination show-last variant where you can click the overflow icon (see https://github.com/eBay/ebayui-core-react/issues/174). This has nothing to do with your PR so you can ignore that part for now.

don-siu:
@abiyasa @HenriqueLimas @darkwebdev 

It should be fixed by now. Please review my PR again. Thank you!

Recording 2022-12-05 at 15 26 36

src/ebay-pagination/helpers.ts
```diff
@@ -1,18 +1,21 @@
...
 ): ItemState[] {
+    const withDots = variant === SHOW_LAST || (variant === OVERFLOW && totalPages > MAX_PAGES)
+    const hasLeadingDots = variant === OVERFLOW && totalPages > MAX_PAGES
     const visibleItems = Math.min(maxVisiblePages, totalPages)
     const startIndexWithoutDots = Math.max(0, selectedPage - Math.ceil((visibleItems - 1) / 2))
     const startIndexWithDots = visibleItems < 4 ? selectedPage :
         Math.max(0, selectedPage - Math.floor((visibleItems - 1) / 2))
     const endIndex = (withDots ? startIndexWithDots : startIndexWithoutDots) + visibleItems
     const closeToEnd = endIndex >= totalPages
+    const closeToFront = selectedPage <= 4
```
HenriqueLimas, line 18: Why `4`? If it is related to line 14, can we add them in a constant?

don-siu: sure will do

don-siu: updated

darkwebdev (approved):

PR #170, opened 2022-11-29
Added id prop to EbayPageNotice

kristopherrollert:
Accessibility Team filed a bug for the below issue.
> **Expected Result**
The value assigned to an id attribute used in ARIA or in form labels must be unique to prevent the second instance from being overlooked by assistive technology.
> **Actual Result**
Document has multiple elements referenced with ARIA with the same id attribute: attention-status

Some eBay applications use multiple of these EbayPageNotice with the same status, so the ids end up being the same. So, to fix this without breaking other applications, I added an optional override to that key.


abiyasa (changes requested):
Thanks a lot for the PR! Do you mind adding a test for this? 

One option is to add a story like:

```jsx
.add(`With id`, () => (<>
  <EbayPageNotice status="confirmation" aria-label="Success" id="main-page-notice">
    <EbayNoticeContent>
      <EbayPageNoticeTitle>Congrats!</EbayPageNoticeTitle>
      <p>You just listed your first item.</p>
    </EbayNoticeContent>
  </EbayPageNotice>
</>))
```

Every storybook will be transformed into snapshot test via storyshot.

Apart from story, all lgtm 👍

abiyasa:
@kristopherrollert  The snapshot test on CI is failing. Could you run `yarn test -u` locally and commit the updated snapshot file?

PR #180, opened 2022-12-15
[Fix] Pagination show-last variant overflow button click disabled

flakronkadriu:
- Fix on dots not clickable for "show-last" variation

Storybook: https://opensource.ebay.com/ebayui-core-react/pagination-clickable-overflow-fix/index.html?path=/story/ebay-pagination--variant-showlast-interactive

### DIFF
--- a/src/ebay-pagination/pagination.tsx
+++ b/src/ebay-pagination/pagination.tsx
@@ -108,7 +108,7 @@ const EbayPagination: FC<PaginationProps> = ({
                 role: isDot ? 'separator' : undefined,
                 key,
                 hide,
-                onPrevious, onSelect, onNext, a11yPreviousText, a11yNextText,
+                onPrevious, onNext, onSelect, a11yPreviousText, a11yNextText,
                 ref: childPageRefs.current[index]
             }
             // include hidden numbers & number of (...)itself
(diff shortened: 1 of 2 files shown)


src/ebay-pagination/pagination.tsx
```diff
@@ -102,11 +103,12 @@ const EbayPagination: FC<PaginationProps> = ({
             const hide = page[index] === 'hidden'
             const newProps = {
                 type, current, disabled, href,
-                children: isDot ? '…' : text,
+                children: isDot ? <EbayIcon name="overflow" height="18px" focusable={false} /> : text,
                 pageIndex: type === 'page' ? pageIndex++ : undefined,
+                role: isDot ? 'separator' : undefined,
                 key,
                 hide,
-                onPrevious, onNext, onSelect, a11yPreviousText, a11yNextText,
```
don-siu, line 109: Not blocker: Is this order matter?

darkwebdev (changes requested):
I works fine, but I think we don't need to introduce `role` when we already have `type`. Just extend this union type with `separator` or `ellipsys` or `dots` etc and add another switch-case to pagination-item.

src/ebay-pagination/pagination.tsx
flakronkadriu: No not really, i can revert this change

darkwebdev (approved):

Issue #171, opened 2022-11-30
Make callbacks consistent


darkwebdev:
## eBayUI Core Marko
For example ebay-menu/ebay-menu-button (parameters passed via an object):
```
onChange (radio): { el, index, checked, originalEvent }
onChange (checkbox): { el, [indexes], [checked], originalEvent }
onChange (not radio or checkbox): { el, index, checked, originalEvent }
onSelect: { el, originalEvent, index }
keydown: { el, index, checked }
```

## eBayUI Core React
Currently (parameters passed via comma-separated arguments):
```
onChange: (index, value, event)
```
I'd suggest to switch from arguments to object like in Marko version and deprecate old variant:
```
onChange: ({ index: number, value: string | number, originalEvent: ChangeEvent })
onChange (checkboxes): ({ indexes: number[], values: string[] | number[], originalEvent: ChangeEvent })
onKeyDown: ({ index: number, key: string, originalEvent: KeyboardEvent })
```
That's not only consistent with marko version, but also makes it easier to introduce new parameters.
For now we can check if there's only one argument which is an object -> treat it the new way, otherwise use deprecated way. So that we don't have to break compatibility yet.

aaron-nance:
I'm definitely for making the API more closely match what Marko has just for the sake of making it simpler to migrate.

abiyasa:
I would against switching from arguments to object. 

While that will make our API similar to Marko but this might cause lots of problem to us:

- I never seen any React UI library which returns 1 argument object for its event handler. From most popular libraries like [MaterialUI](https://mui.com/), [SemanticUI](https://react.semantic-ui.com/), [ChakraUI](https://chakra-ui.com/), [ReachUI](https://reach.tech/), and [Microsoft FluentUI](https://developer.microsoft.com/en-us/fluentui#/controls/web), all of them are using multiple arguments on its event handler.
- Different idiom: We're using React and there are many things how it's different than Marko. Marko does event emitter with `emit()` while in React, we're calling the event handler function directly. The event that emitted by Marko component will be propagated automatically to its parent components, something that React does not have. So while we're always trying to use the same API, there will be some difference due to React vs Marko idiom. 
- We will have more problems with TypeScript. For example, our EbayButton extends some props from React HTMLButtonElement which has `onChange(e: SyntheticMouseEvent)`, or other component `onChange(e: ChangeEvent<HTMLInputElement>)`. If we switch to object, we need to omit and redefine all React HTML element's event handlers on our component.
[...]

PR #181, opened 2022-12-21
EbayMenu: Improve callbacks

darkwebdev:
- Implements #171 for EbayMenu
- Add callback tests
- Add story actions to see those callbacks
- Follow marko API

I'm making this PR against `improved-callbacks` branch, cause those callbacks are incompatible with old ones and will break the code that uses them. We'll bump major version when we merge that branch.

storybook: https://opensource.ebay.com/ebayui-core-react/EbayMenu-callbacks/index.html?path=/story/ebay-menu--default

### DIFF
--- a/src/ebay-menu/types.ts
+++ b/src/ebay-menu/types.ts
@@ -1,2 +1,28 @@
+import { ComponentProps, Ref } from 'react'
+import { EbayEventHandler, EbayKeyboardEventHandler } from '../common/event-utils/types'
+
 export type EbayMenuType = 'radio' | 'checkbox'
 export type EbayMenuPriority = 'primary' | 'secondary' | 'none'
+
+type SpanProps = Omit<ComponentProps<'span'>, 'onKeyDown' | 'onChange'>
+type SelectProps = {
+    index: number;
+    checked: number[];
+}
+type ChangeProps = SelectProps & {
+    indexes: number[];
+    checkedValues: string[];
+}
+export type EbayMenuKeyDownEventHandler = EbayKeyboardEventHandler<HTMLElement, SelectProps | ChangeProps>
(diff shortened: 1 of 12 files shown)


src/ebay-menu/types.ts
```diff
@@ -1,2 +1,28 @@
...
+type SpanProps = Omit<ComponentProps<'span'>, 'onKeyDown' | 'onChange'>
+type SelectProps = {
+    index: number;
+    checked: number[];
+}
+type ChangeProps = SelectProps & {
+    indexes: number[];
+    checkedValues: string[];
+}
+export type EbayMenuKeyDownEventHandler = EbayKeyboardEventHandler<HTMLElement, SelectProps | ChangeProps>
```
darkwebdev, line 16: Exporting handler types here...

src/ebay-menu/types.ts
```diff
@@ -1,2 +1,28 @@
...
 export type EbayMenuType = 'radio' | 'checkbox'
 export type EbayMenuPriority = 'primary' | 'secondary' | 'none'
+
+type SpanProps = Omit<ComponentProps<'span'>, 'onKeyDown' | 'onChange'>
+type SelectProps = {
+    index: number;
+    checked: number[];
+}
+type ChangeProps = SelectProps & {
+    indexes: number[];
```
flakronkadriu, line 13: why do we return the indexes as well?

flakronkadriu (commented):
Looks good to me 👍🏻

src/ebay-menu/types.ts
darkwebdev: I guess the idea was to return indexes that are checked in case of `radio` or `checkbox` type of the EbayMenu.
That's the original EbayUI API that was implemented in Marko.

PR #183, opened 2022-12-28
EbayMenuItem supports `badgeAriaLabel`

darkwebdev:


src/ebay-menu/menu-item.tsx
```diff
@@ -33,16 +35,16 @@ const EbayMenuItem: FC<MenuItemProps> = ({
 
     return (
         <div
+            aria-label={badgeAriaLabel}
             {...rest}
             ref={ref}
             className={classNames(className, 'menu__item', hasBadge && 'menu__item--badged')}
             role="menuitem"
-            aria-checked={checked}
+            aria-checked={checked || undefined}
```
flakronkadriu, line 43: maybe `false` instead of `undefined`? Or assign `checked` a default value of `false`? I assume since the value is optional, it means it's undefined by default.

darkwebdev: `checked` comes as a prop from EbayMenu, I can move `|| undefined` there :)

darkwebdev: fixed

PR #185, opened 2023-01-02
[FEAT] EbayVideo: new component

darkwebdev:
Marko version: [source](https://github.dev/eBay/ebayui-core/blob/master/src/components/ebay-video/README.md), [storybook](https://opensource.ebay.com/ebayui-core/?path=/story/media-ebay-video--standard)

Storybook: https://opensource.ebay.com/ebayui-core-react/ebay-video/index.html?path=/story/ebay-video--default

We don't have many tests, only snapshots cause it's hard to test this player, it doesn't work in react testing library. Marko version has no tests at all.

Props that are unsupported yet:
- cdnUrl
- cssUrl
- cdnVersion

And we don't load the player from our CDN yet, we'll optimize later.


src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+    const handleOnPause = () => {
+        // On IOS, the controls force showing up if the video exist fullscreen while playing.
+        // This forces the controls to always hide
+        videoRef.current.controls = false
+    }
+
+    return (
+        <div
+            style={{ width: `${width}px`, height: `${height}px` }}
```
HenriqueLimas, line 207: With and height are not required, we are always passing `undefinedpx`.

src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+                    style={{ width: `${width}px`, height: `${height}px` }}
+                    poster={thumbnail}
+                    muted={muted || false}
+                    onPlaying={handlePlaying as any}
+                    onPause={handleOnPause as any}
+                    onVolumeChange={handleVolumeChange}
+                    {...rest}
+                >
+                    {sources.map(source => {
+                        <source {...source} />
```
HenriqueLimas, line 237: Missing `key` attribute`

src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+    ...rest
+}) => {
+    const [loaded, setLoaded] = useState<boolean>()
+    const [buffering, setBuffering] = useState<boolean>()
+    const [playing, setPlaying] = useState<boolean>()
+    const [failed, setFailed] = useState<boolean>()
+
+    const containerRef = useRef<HTMLDivElement>(null)
+    const videoRef = useRef<HTMLVideoElement>(null)
+    const playerRef = useRef<any>(null)
```
HenriqueLimas, line 66: We shouldn't use `any`, we should define the proper type

src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+    }
+
+    const loadSource = (index = 0) => {
+        if (!sources.length || !playerRef.current) return
+
+        setLoaded(false)
+        playerRef.current
+            .load(sources[index]?.src)
+            .then(() => {
+                setFailed(false)
```
HenriqueLimas, line 84: Before calling `setFailed` we need to check that the component is not unmounted, React will thrown an error in the console in case a change in state is executed when the element is unmounted. We can create a `safeSetFailed()` that runs only when the element is present:

```jsx
const isUnmounted = useRef(false)

useEffect(() => {
   return () => {
     isUnmounted.current = true
   }
}, [])

const safeSetFailed = value => {
   if (!isUnmounted.current) {
      setFailed(value)
   }
}
```

src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+            })
+            .catch(err => {
+                console.error(err)
+                switch (err.code) {
+                    case ERROR_ANOTHER_LOAD:
+                        return
+                    case ERROR_NO_PLAYER:
+                        setTimeout(() => loadSource(index), 0)
```
HenriqueLimas, line 92: Any reason we are using setTimeout? Should we use `requestAnimateFrame` or `requestidlecallback`? We can add a comment here explaining why setTimeout

src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+        setLoaded(false)
+        playerRef.current
+            .load(sources[index]?.src)
+            .then(() => {
+                setFailed(false)
+            })
+            .catch(err => {
+                console.error(err)
+                switch (err.code) {
```
HenriqueLimas, line 88: Should we `setFailed(true)` in the `catch`?

src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+        })
+
+        const { Report, TextSelection } = customControls(onReport)
+        shaka.ui.Controls.registerElement('report', new Report.Factory(reportText))
+        shaka.ui.Controls.registerElement('captions', new TextSelection.Factory())
+
+        loadSource()
+        hideSpinner(container)
+
+        // return () => {
```
HenriqueLimas, line 145: Shouldn't we destroy the elements on unmounted?

src/ebay-video/video.tsx
```diff
@@ -0,0 +1,270 @@
...
+    volumeSlider?: boolean;
+    playView?: VideoPlayView;
+    cdnUrl?: string;
+    cssUrl?: string;
+    cdnVersion?: string;
+    a11yLoadText: string;
+    a11yPlayText: string;
+    errorText: string;
+    reportText: string;
+    onLoadError?: (err: any) => void;
```
HenriqueLimas, line 29: We shouldn't use `any` on new components

src/ebay-video/video.tsx
darkwebdev: we either:
- ignore if another load in progress
- retry if player not loaded yet
- load next available video from playlist
- `setFailed(true)` inside `handleError()` in all other cases

src/ebay-video/video.tsx
darkwebdev: we get errors when we do that, needs further investigation

darkwebdev:
created issue  #188 from comments

Issue #188, opened 2023-01-04
EbayVideo: improvements


darkwebdev:
Fixes and improvements from review and feedback:
- [x] missing `key` attribute in `sources.map`
- [x] fix `undefinedpx` when `width`/`height` props are not passed
- [x] get rid of `any` types
- [x] allow to hide report button
- [x] hide menu button if menu is empty
- [x] show captions button instead of menu?
- [ ] support partial loading from CDN to speedup player initialization and video load
- [ ] can we get rid of `setTimeout` in favor of `requestAnimateFrame` or `requestidlecallback`? Or at least explain why we use it?
- [ ] make sure all components are destroyed on unmount and no errors appear. For example:

Before calling setFailed we need to check that the component is not unmounted, React will thrown an error in the console in case a change in state is executed when the element is unmounted. We can create a safeSetFailed() that runs only when the element is present:
```js
const isUnmounted = useRef(false)

useEffect(() => {
   return () => {
     isUnmounted.current = true
   }
}, [])

const safeSetFailed = value => {
   if (!isUnmounted.current) {
      setFailed(value)
   }
}
```

PR #189, opened 2023-01-04
EbayVideo: improvements

darkwebdev:
implements #188 
- add missing key attribute in sources.map
- fix `undefinedpx` when width/height props are not passed
- allow to hide report button
- show captions button instead of menu?
- get rid of `any` type

### DIFF
--- a/src/ebay-video/__tests__/index.stories.tsx
+++ b/src/ebay-video/__tests__/index.stories.tsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react'
 import { Meta } from '@storybook/react'
-import { EbayButton, EbayVideo, EbayVideoSource } from '../../index'
+import { EbayButton, EbayVideo, EbayVideoProps, EbayVideoSource } from '../../index'
 import { action } from '../../../.storybook/action'
 
 export default {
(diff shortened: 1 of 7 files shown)


src/ebay-video/__tests__/index.stories.tsx
```diff
@@ -1,18 +1,21 @@
 import React, { useState } from 'react'
 import { Meta } from '@storybook/react'
-import { EbayButton, EbayVideo, EbayVideoSource } from '../../index'
+import 'shaka-player/dist/controls.css'
```
HenriqueLimas, line 3: We need to add this info in the readme also that applications needs to import this css

src/ebay-video/const.ts
```diff
@@ -11,8 +11,8 @@ export const defaultVideoConfig = {
         'spacer',
         'mute',
         'report',
-        'fullscreen',
-        'overflow_menu'
-    ],
-    overflowMenuButtons: ['captions']
+        'captions',
+        // 'quality',
```
HenriqueLimas, line 15: Do we need this comment?

darkwebdev: added comment for this

src/ebay-video/__tests__/index.stories.tsx
darkwebdev: added

HenriqueLimas (approved):

PR #190, opened 2023-01-05
FIX: Inline notice - Add role="region"

praveengh1:


src/ebay-inline-notice/inline-notice.tsx
```diff
@@ -42,7 +43,10 @@ const EbayInlineNotice: FC<Props> = ({
     const isGeneral = status === `general`
 
     return (
-        <div className={classNames(className, `inline-notice ${!isGeneral ? `inline-notice--${status}` : ``}`)}>
+        <div
+            {...rest}
+            className={classNames(className, `inline-notice ${!isGeneral ? `inline-notice--${status}` : ``}`)}
```
darkwebdev, line 48:
```suggestion
            className={classNames(className, 'inline-notice', {[ `inline-notice--${status}`]: !isGeneral )}
```

src/ebay-inline-notice/inline-notice.tsx
```diff
@@ -6,7 +6,7 @@ import { findComponent } from '../common/component-utils'
 import { EbayIcon, Icon } from '../ebay-icon'
 import { NoticeStatus } from './types'
 
-type Props = {
+type Props = React.HTMLProps<HTMLDivElement> & {
```
darkwebdev, line 9:
```suggestion
type Props = React.ComponentProps<'div'> & {
```

darkwebdev (dismissed):

darkwebdev:
Hey @praveengh1 ! Thanks for the PR, looks good, except for a couple of places where I've suggested to use a code a bit differently...

src/ebay-inline-notice/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -131,7 +131,7 @@ exports[`Storyshots ebay-inline-notice Notice toggle 1`] = `
     Hide Notice
   </button>
   <div
-    class="inline-notice inline-notice--confirmation"
+    class="inline-notice inline-notice--\${status}"
```
HenriqueLimas, line 134: This should be `confirmation` not `${status}` this is because of using single quote instead of backtick (comment bellow)

src/ebay-inline-notice/inline-notice.tsx
HenriqueLimas: @darkwebdev could you give a reason why to make this change just to understand. We need to have a consistency between components, (the HTMLProps was inspired by [page-notice](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-page-notice/page-notice.tsx#L6)). Also all the notice utils uses it HTMLProps

HenriqueLimas (approved):

PR #194, opened 2023-01-11
[FEAT] EbayListboxButton: Adding missing tabIndex functionality from marko component

marcial199545:
Adding missing functionality to have dynamic `tabIndex` with value `0`when list is open and `-1` when it is closed.
https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-listbox/index.marko#:~:text=tabindex%3D(option.disabled%20%3F%20%2D1%20%3A%20option.tabindex)

### DIFF
--- a/src/ebay-listbox-button/listbox-button.tsx
+++ b/src/ebay-listbox-button/listbox-button.tsx
@@ -61,6 +61,8 @@ const ListboxButton: FC<EbayListboxButtonProps> = ({
     const [expanded, setExpanded] = useState(false)
     // Additional flag to avoid multiple re-render when users tries to open and close
     const [optionsOpened, setOptionsOpened] = useState(false)
+    // variable tabIndex
+    const [listContainerTabIndex, setListContainerTabIndex] = useState(-1)
 
     useEffect(() => {
         setSelectedOption(selectedOptionFromValue)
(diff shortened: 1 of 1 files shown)


darkwebdev:
Hello! Thanks for the PR! I'm trying to understand which problem are we solving here?
[Skin](https://opensource.ebay.com/skin/#listbox-button) always has `tabIndex=0`
[Marko](https://opensource.ebay.com/ebayui-core/?path=/story/buttons-ebay-listbox-button--standard) switches between 0 and -1
But is there a difference from a11y perspective? I thought that if menu is hidden then tabIndex doesn't matter...

src/ebay-listbox-button/listbox-button.tsx
```diff
@@ -235,7 +244,7 @@ const ListboxButton: FC<EbayListboxButtonProps> = ({
                     <div
                         className="listbox-button__options"
                         role="listbox"
-                        tabIndex={0}
+                        tabIndex={listContainerTabIndex}
```
darkwebdev, line 247:
```suggestion
                        tabIndex={expanded ? 0 : -1}
```
and you don't need any more changes from above

darkwebdev (changes requested):

marcial199545:
retest

marcial199545:
closing PR, to reopen it with changes proposed by [Timur Manyanov](https://github.com/darkwebdev)

PR #195, opened 2023-01-11
FIX: Add missing types on components for React 18

HenriqueLimas:
- Define props for components that are used only for extracting props on runtime (NoticeContent, TooltipHost, TooltipContent)

This is because for React 18 `children` prop is required to be defined when used with children in the application.


darkwebdev:
Do we even need `children` in Prop types? I think `FC` automatically has `children` inside.

darkwebdev (approved):

PR #201, opened 2023-01-19
Remove usage of global imports

darkwebdev:
Now it should be safe to remove global `index.ts` in next major release.

How I've tested:
- removed global index.ts
- yarn build
- yarn test
- yarn storybook
- 👍

### DIFF
--- a/src/ebay-menu/__tests__/index.spec.tsx
+++ b/src/ebay-menu/__tests__/index.spec.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
-import { render, fireEvent, screen } from '@testing-library/react'
-import { act } from '@testing-library/react-hooks'
+import { render, fireEvent } from '@testing-library/react'
 import { initStoryshots } from '../../../config/jest/storyshots'
 import { EbayMenu, EbayMenuItem } from '../index'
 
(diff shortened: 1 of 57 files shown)


HenriqueLimas (commented):
We need to add BREAKING in the title/comment. Should we move to the branch that we are doing the callbacks?

darkwebdev:
@HenriqueLimas no, this PR only removes using of global imports inside our project. It doesn't break anything.

src/ebay-menu/__tests__/index.spec.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
-import { render, fireEvent } from '@testing-library/react'
+import { render, fireEvent, screen } from '@testing-library/react'
+import { act } from '@testing-library/react-hooks'
```
HenriqueLimas, line 3: Act and screen are not being used do we need to import it?

darkwebdev: @HenriqueLimas thanks, removed. I've created an issue regarding that problem recently: #202

HenriqueLimas (approved):

PR #209, opened 2023-01-30
FIX: Remove tooltip role from infotip

### LINKED ISSUE #208: ebay-infotip  : A11y issue - To remove the role "tooltip" for the infotip component
Issue: We have received a A11y issue saying that role "tooltip" is not the right role assigned.

```There are dialogs spawned by info buttons that have a tooltip role assigned.

A tooltip is a contextual text bubble that displays a description for an element that appears on pointer hover or keyboard focus.

In this case these information containers function as dialogs, as they are only displayed after activating the triggering info buttons and they need to be closed by using the close button in the modal or tabbing back to the triggering button and activating it again to get the modal closed.
```

The A11y team asked to raise it as issue to remove the role for the infotip component. This fix will help us to fix all the A11y issue using infotip component used in the page.

https://github.com/eBay/ebayui-core-react/blob/main/src/common/tooltip-utils/tooltip-content.tsx#L38
image

HenriqueLimas:
Following Marko [implementation](https://github.com/eBay/ebayui-core/blob/master/src/components/components/ebay-tooltip-overlay/index.marko#L28)

- infotip -> remove `role`
- tooltip -> role="tooltip"
- tourtip -> role="region"

Fixes #208

### DIFF
--- a/src/common/tooltip-utils/constants.ts
+++ b/src/common/tooltip-utils/constants.ts
@@ -1,9 +1,15 @@
-import { PointerDirection } from './types'
+import { PointerDirection, TooltipType } from './types'
 
 export const DEFAULT_POINTER_DIRECTION: PointerDirection = 'bottom'
 
+export const TYPE_ROLES: Record<TooltipType, string> = {
+    tooltip: 'tooltip',
+    tourtip: 'region',
+    infotip: ''
+}
+
 export const POINTER_STYLES: { [key in PointerDirection]: any } = {
-    'left': {
+    left: {
         transform: 'translateX(16px) translateY(-50%) scale3d(1,1,1)',
         left: '100%',
         right: 'auto',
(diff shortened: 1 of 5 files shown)


shpandian (approved):
LGTM

src/common/tooltip-utils/constants.ts
```diff
@@ -1,9 +1,15 @@
-import { PointerDirection } from './types'
+import { PointerDirection, TooltipType } from './types'
 
 export const DEFAULT_POINTER_DIRECTION: PointerDirection = 'bottom'
 
+export const TYPE_ROLES: Record<TooltipType, string> = {
+    tooltip: 'tooltip',
+    tourtip: 'region',
+    infotip: ''
```
shpandian, line 8: do we need to say ""? it is as good as having null right?

HenriqueLimas: This is for typescript not complaining 😢

PR #211, opened 2023-02-17
Add buttonPosition attribute check for close icon-button - Dialog

vshrivastav13:
Change:

This PR adds a check to hide the close button on drawer-dialog ( `X`  button ) when `noHandle` attribute is set to true.

image

### DIFF
--- a/src/ebay-dialog-base/components/dialogBase.tsx
+++ b/src/ebay-dialog-base/components/dialogBase.tsx
@@ -57,7 +57,7 @@ export const DialogBase: FC<DialogBaseProps<HTMLElement>> = ({
     mainId,
     top,
     header,
-    buttonPosition = 'left',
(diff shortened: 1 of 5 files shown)


HenriqueLimas (changes requested):
It seems noHandle is used for hiding the expand button already. In order to be aligned with marko we need to:
- On dialogBase, we need to set default buttonPosition as `right`, same as marko [default](https://github.com/eBay/ebayui-core/blob/master/src/components/components/ebay-dialog-base/index.marko#L39)
- We can remove buttonPosition from the [drawer](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-drawer-dialog/components/drawer.tsx#L103)

HenriqueLimas (approved):

PR #216, opened 2023-03-03
[#212, #213] Make a11y labels optional

### LINKED ISSUE #212: EbayDrawerDialog: a11yMinimizeText and a11yMaximizeText should be optional
a11yMinimizeText  and a11yMaximizeText should be optional. Currently they are set as required, but not all drawers use the maximize and minimize feature. An example would be a drawer that is used as a confirmation modal so the controls to dismiss the drawer will be part of the message on the control.

### LINKED ISSUE #213: EbayConfirmDialog: a11yCloseText should be optional
Currently a11yCloseText is required, but confirmation modals may not have a close button.

darkwebdev:
fixes #212 
fixes #213


aaron-nance (approved):
Looks good. Thanks Timur!

Issue #217, opened 2023-03-21
Skin 16.0.0 changes 


montoya332:
16.0.0 Roadmap: 
https://github.com/orgs/eBay/projects/8/views/2

Callouts : 

- Icon changes including program badges : 
  - icons
    - https://github.com/eBay/skin/pull/2016
    - https://github.com/eBay/ebayui-core/issues/1869
  - bages:
    - https://github.com/eBay/skin/pull/2013
    - https://github.com/eBay/ebayui-core/issues/1871

darkwebdev:
released in v4.0.0

PR #219, opened 2023-04-04
[#217] Upgrade skin to 16

darkwebdev:
- Skin 16
- Remove EbayProgramBadge component
- Update icons
- Update icon generating scripts
- Fix all icon names in components
- Update tests/snapshots

Storybook: https://opensource.ebay.com/ebayui-core-react/skin16/?path=/story/ebay-icon--all-icons


src/ebay-cta-button/cta-button.tsx
```diff
@@ -31,7 +31,7 @@ const EbayCtaButton: FC<Props> = ({
         <a {...rest} className={className} ref={forwardedRef}>
             <EbayButtonCell type="cta">
                 <span>{children}</span>
-                <EbayIcon name="cta" width={8} height={8} />
+                <EbayIcon name="arrowRight24" width={8} height={8} />
```
aaron-nance, line 34: Would `aarowRight12` be a better choice here and then remove the`width` and `height` attributes.

src/ebay-pagination/pagination.tsx
```diff
@@ -105,7 +105,7 @@ const EbayPagination: FC<PaginationProps> = ({
             const newProps = {
                 current, disabled, href,
                 type: isSeparator ? 'separator' : type,
-                children: isDot ? <EbayIcon name="overflow" height="18px" focusable={false} /> : text,
+                children: isDot ? <EbayIcon name="overflowHorizontal24" height="18px" focusable={false} /> : text,
```
aaron-nance, line 108: Would `overflowHorizontal16` or `overflowHorizontal20` work here? Hopefully we can avoid overriding sizes.

aaron-nance (commented):
Had a couple comments about places where we were overriding the sizes of icons. Other than these very minor things this all looks excellent.

shpandian:
> You have successfully added a new ESLint configuration `.github/workflows/pr.yml:analyze`. As part of the setup process, we have scanned this repository and found no existing alerts. In the future, you will see all code scanning alerts on the repository [Security tab](/eBay/ebayui-core-react/security/code-scanning?query=pr%3A219+tool%3AESLint+is%3Aopen).

@darkwebdev is this something we explicitly integrated? wondering what sort of basic rules, and if it could be applied to our internal repos as well?

src/ebay-cta-button/cta-button.tsx
shpandian: am also wondering if we need to render 24 one, shouldn't it to be expected to be same size, sorry - trying to understand height and width on top of it again.

The references am seeing from skin site is like that

https://opensource.ebay.com/skin/#icon

scripts/update-icons-helpers.js
```diff
@@ -13,21 +13,23 @@ function parseSVG(skinIconsFile) {
 
 function getIconKeys(icons = []) {
     const symbolIds = icons.map(symbol => symbol.attributes.id)
+    const skipped = symbolIds.filter(id => !id.startsWith('icon-'));
+
+    if (skipped.length) {
+        console.log(`Skipping ${skipped.length} icons: ${skipped.join(', ')}.`)
```
shpandian, line 19: do we need console log?

src/ebay-cta-button/cta-button.tsx
darkwebdev: Thanks for paying attention to that part, I need to remove `width: 8, height: 8`, cause they don't work anyway. Yeah we can use arrowRight12, cause it's overridden by CSS: 
image

src/ebay-pagination/pagination.tsx
darkwebdev: you're right, missed those override attempts, we don't need them.

scripts/update-icons-helpers.js
darkwebdev: this is just for icons generating script, debug info for devs, so that they notice if something's wrong with the icons...

PR #168, opened 2022-11-23
EbaySplitButton

darkwebdev:
Storybook:
https://opensource.ebay.com/ebayui-core-react/ebay-split-button/index.html?path=/story/ebay-split-button--default

Original Marko split-button:
https://opensource.ebay.com/ebayui-core/?path=/story/buttons-ebay-split-button--standard

PR #231, opened 2023-06-01
FEAT: adding ebay menu button component and ebay menu with existing callback

shpandian:
This PR is first attempt to accommodate the ebay-split-button from this PR - https://github.com/eBay/ebayui-core-react/pull/168

1. adding EbayMenuButton component with right callback support, icon name change
2. update on EbayMenu to the right callback API

### DIFF
--- a/src/common/event-utils/types.ts
+++ b/src/common/event-utils/types.ts
@@ -1,6 +1,3 @@
-import { ChangeEvent, KeyboardEvent, MouseEvent, FocusEvent, SyntheticEvent } from 'react'
-
-
 type ModifierKeys = 'Alt' | 'AltGraph' | 'Control' | 'Shift' | 'CapsLock' | 'Meta'
 // | 'Fn' | 'FnLock' | 'Hyper' | 'NumLock' | 'ScrollLock' | 'Super' | 'Symbol' | 'SymbolLock'
 type NavigationKeys = 'ArrowDown' | 'ArrowLeft' | 'ArrowRight' | 'ArrowUp' | 'Enter' | 'Tab' | ' ' | 'Escape'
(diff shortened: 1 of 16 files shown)


src/common/event-utils/types.ts
```diff
@@ -9,3 +12,16 @@ type NavigationKeysEdge = 'Down' | 'Left' | 'Right' | 'Up' | 'Esc'
 // 'Separator' | '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'
 
 export type Key = ModifierKeys | NavigationKeys | NavigationKeysEdge
+
+type BaseEventHandler<E extends SyntheticEvent<any>, P> = (event: E, props?: P) => void;
+
+export type EbayEventHandler<ElementType = Element, PropsObject = Record<string, any>> =
```
shpandian, line 18: i am not using this for now -- just picked from the other branch.

src/ebay-menu-button/menu-button.tsx
```diff
@@ -0,0 +1,176 @@
...
+            onClick(e)
+        },
+        ...rest
+    }
+
+
+    return (
+        <span className={wrapperClasses}>
+            {variant === 'overflow' ?
+                <EbayIconButton icon="overflowHorizontal16" {...buttonProps} /> :
```
shpandian, line 132: is this the right one? overflowHorizontal16 used in place of overflow

src/ebay-menu/menu.tsx
```diff
@@ -87,7 +88,7 @@ const EbayMenu: FC<Props> = ({
     }, [checkedIndexes])
 
     return (
-        <span {...rest} className={classNames(className, 'menu')}>
+        <div {...rest} className={classNames(className, 'menu')}>
```
HenriqueLimas, line 91: We need to be able to support both div and span. By default it will be a span, but we pass the other `baseEl` element as a prop. (We do something similar on  [base dialog](https://github.com/eBay/ebayui-core-react/blob/bf552d6611c82d22368f3cbdfb4b3ffe43b9c5d1/src/ebay-dialog-base/components/dialogBase.tsx#L53))

src/ebay-menu-button/menu-button.tsx
HenriqueLimas: Marko [version](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu-button/index.marko#L87) uses overflowVertical16

shpandian: updated
image

src/ebay-menu/menu.tsx
shpandian: with span remember we noticed that there was display:none from menu.css

src/common/event-utils/types.ts
shpandian: removed

shpandian: ah! i was removing it from menu comp types file. :)

shpandian: https://github.com/eBay/ebayui-core-react/pull/231/commits/1b64e1aed8d2ffa244686fbc16c0d988ec03fa8e

src/ebay-menu/menu.tsx
shpandian: added support for span and div
https://github.com/eBay/ebayui-core-react/pull/231/commits/a53baf4a9ae5347f602582c94fbeb18c35baef12

shpandian:
@HenriqueLimas i have addressed the comments as we discussed, please take another look - Thanks!

HenriqueLimas (approved):

PR #235, opened 2023-07-05
Fixing focus issue on drop down components

rpennamareddy:
Focus is now returned to the drop down element on changing country or region


HenriqueLimas:
The issue is not on the library but how the application handles focus internally. On storybook this error is not visible so the change needs to be investigated and fixed in the application

PR #239, opened 2023-07-28
FIX: EbayInfotip - move drawer as a sibling of the infotip

### LINKED ISSUE #238: EbayInfotip: Variant 'modal' layout issue on close button
# Bug Report

## Description

Use ebay-infotip with variant `modal` ([storybook](https://opensource.ebay.com/ebayui-core-react/main/index.html?path=/story/ebay-infotip--modal))

The close button of the drawer dialog is rendering incorrectly.

The reason is that the `.drawer-dialog` is being rendered as a child of `.infotip` and there is a `.infotip .icon-btn` that impacts in the rendering. The `.drawer-dialog` should be rendered as a sibling of `.infotip` the same as [skin version](https://opensource.ebay.com/skin/#infotip).

## Screenshots

Screen Shot 2023-07-26 at 3 43 34 PM

HenriqueLimas:
Fixes #238 

- Move the drawer dialog as a sibling of the infotip in that way styles are not conflicting between infotip and drawer styles

Screen Shot 2023-07-28 at 10 00 13 AM

### DIFF
--- a/src/common/tooltip-utils/tooltip.tsx
+++ b/src/common/tooltip-utils/tooltip.tsx
@@ -29,7 +29,7 @@ const Tooltip: FC<TooltipProps> = ({
     const host = cloneElement(originalHostComponent, {
         className: `${type}__host`,
         'aria-expanded': isExpanded,
-        'aria-describedby': content?.props.id,
+        'aria-describedby': content?.props?.id,
         ...originalHostComponent.props
     })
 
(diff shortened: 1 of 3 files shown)


src/common/tooltip-utils/tooltip.tsx
```diff
@@ -26,14 +26,10 @@ const Tooltip: FC<TooltipProps> = ({
...
 
-    if (!content) {
-        throw new Error(`Tooltip: Please use a component that defines the content of the tooltip`)
-    }
-
     const host = cloneElement(originalHostComponent, {
         className: `${type}__host`,
         'aria-expanded': isExpanded,
-        'aria-describedby': content.props.id,
+        'aria-describedby': content?.props.id,
```
montoya332, line 32: `content?.props?.id`
curious why we dont add check to props aswell

HenriqueLimas: We expect that props is always there by default. But let me add it a check there to just in case.

montoya332 (approved):
LGTM

PR #244, opened 2023-09-07
FIX: Update carousel control icons with latest skin guidelines

HenriqueLimas:
- Use chevron 12 instead of 24 as per skin


flakronkadriu (approved):
Legit 👍🏻

PR #246, opened 2023-09-08
FEAT: Add support for previous button on lightbox

HenriqueLimas:
- Add new `EbayDialogPreviousButton` component to be rendered on lightbox dialogs
- Defaults icon to chevronLeft16 (different then marko that doesn't do any defaulting)
- Upgrade skin to 16.6.0

Screen Shot 2023-09-08 at 4 49 47 PM


src/ebay-dialog-base/dialog-base-with-state.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { Children, ReactElement } from 'react'
 import { RemoveScroll } from 'react-remove-scroll'
+import EbayDialogPreviousButton from './components/dialog-previous-button'
```
flakronkadriu, line 3: maybe import it from index as well?

flakronkadriu:
LGTM 👍🏻

src/ebay-dialog-base/dialog-base-with-state.tsx
HenriqueLimas: Yep, thanks, updated

PR #249, opened 2023-09-28
Fix EbayListboxButton console warning

rglombovsky-ebay:
ListboxButton is using a hidden uncontrolled select so it behaves as a read-only field, therefore it is not possible to use 'value'.

image

### DIFF
--- a/src/ebay-listbox-button/listbox-button.tsx
+++ b/src/ebay-listbox-button/listbox-button.tsx
@@ -286,7 +286,7 @@ const ListboxButton: FC<EbayListboxButtonProps> = ({
                         {updatelistBoxButtonOptions}
                     </div>
                 </div>}
-            <select hidden className="listbox-button__native" name={name} value={selectedOption.props.value}>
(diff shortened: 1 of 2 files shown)


src/ebay-listbox-button/listbox-button.tsx
```diff
@@ -286,7 +286,7 @@ const ListboxButton: FC<EbayListboxButtonProps> = ({
                         {updatelistBoxButtonOptions}
                     </div>
                 </div>}
-            <select hidden className="listbox-button__native" name={name} value={selectedOption.props.value}>
+            <select hidden className="listbox-button__native" name={name} defaultValue={selectedOption.props.value}>
```
HenriqueLimas, line 289: The select value needs to be in sync with the selectedOption state variable. Unfortunately we can't use defaultValue since it doesn't update. We should use `readOnly` instead but select elements doesn't support it. I think we should use an `input type="text" hidden` with `readOnly` instead here.

HenriqueLimas (changes requested):
Thank you very much for the PR!

PR #252, opened 2023-10-03
Fix EbayListboxButton console warning

rglombovsky-ebay:
This is a suggestion on a better way of fixing the warning problem with the hidden controlled select.

Another solution is just to have the onChange with a function that does nothing, but I'd say it's better to just keep the same behaviour as the main select.


darkwebdev:
Thanks for the PR @rglombovsky-ebay . Do you mean this warning?
image

rglombovsky-ebay:
@darkwebdev Yes indeed!

darkwebdev (approved):

PR #254, opened 2023-10-09
feat: added a closeButtonClass prop for base dialog

oxala:
To mirror the ebayui-core-marko dialog: https://github.com/eBay/ebayui-core/blob/master/src/components/components/ebay-dialog-base/index.marko#L54


darkwebdev:
Thanks for the PR!
Could you please also update readme?

darkwebdev (approved):

PR #264, opened 2023-10-19
FIX: Use correct aria attribute names in Tourtip

rglombovsky-ebay:
There is currently a console error when using Tourtips. It's due to the wrong naming for aria attributes. According to React's documentation, aria attributes should not be camel case (https://react.dev/reference/react-dom/components/common#common-props)

Screenshots

Error:
image
image

Fix:
image


darkwebdev:
thank you!

PR #266, opened 2023-10-31
fix onBlur floating label logic

montoya332:
set to false only if there is no value in element 
https://opensource.ebay.com/ebayui-core-react/main/index.html?path=/story/ebay-select--floating-label


HenriqueLimas (commented):
Can we add unit test for this use case

HenriqueLimas (approved):

PR #269, opened 2023-11-03
FEAT: Update skin and icons

HenriqueLimas:
- Update to skin 16.8.0
- Update icons
- Update icon script logic to consider other elements other then `symbols`
- Change EbaySvg from display none to position absolute so `<defs>` are shown properly

https://opensource.ebay.com/ebayui-core-react/update-icons/index.html?path=/story/ebay-icon--all-icons


scripts/update-icons.js
```diff
@@ -3,19 +3,19 @@
...
-const skinSVGSymbols = parseSVG(skinIconsFile);
+const skinSVG = parseSVG(skinIconsFile)
 
-console.log(`Found ${skinSVGSymbols.length} icons in Skin.`)
-
-saveSvgIcons(skinSVGSymbols, ebaySvgFile)
+saveSvgIcons(skinSVG, ebaySvgFile)
 
+const skinSVGSymbols = parseSVGSymbols(skinIconsFile)
+console.log(`Found ${skinSVGSymbols.length} icons in Skin.`)
```
shpandian, line 18: not part of this PR, but curious  - do we have any significance in this console log?

shpandian (approved):
LGTM

scripts/update-icons.js
HenriqueLimas: I think it just show as a feedback that the commend works

Issue #255, opened 2023-10-11
Switch to Node 18


darkwebdev:
- update .nvmrc
- test that everything works

PR #271, opened 2023-11-13
[#255] Switch to Node 18

darkwebdev:


HenriqueLimas (commented):
We need to also update the github action to use the legacy openssl https://github.com/eBay/ebayui-core-react/blob/main/.github/actions/storybook/action.yml#L15C31-L15C31

HenriqueLimas (approved):

PR #272, opened 2023-11-14
FIX: Add types on tests and fix type definitions

HenriqueLimas:
- Add a new type check config that checks types for test folder also
- Fix types for EbayBadge, EbayInfotipHost
- Fix test types


montoya332 (approved):
thank you

PR #283, opened 2023-11-15
FEAT: EbayMenuButton: support prefixId/prefixLabel

darkwebdev:
Storybook: [use case 1](https://opensource.ebay.com/ebayui-core-react/menu-button-prefix/index.html?path=/story/ebay-menu-button--with-custom-label) [use case 2](https://opensource.ebay.com/ebayui-core-react/menu-button-prefix/index.html?path=/story/ebay-menu-button--single-select-menu-button-collapse-on-click)


src/ebay-menu-button/menu-button.tsx
```diff
@@ -49,6 +48,8 @@ const EbayMenuButton: FC<Props> = ({
     checked,
     collapseOnSelect,
     a11yText,
+    prefixId,
```
HenriqueLimas, line 51: Marko [implementation](https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-menu-button/index.marko#L96) is a little bit different, should we use that same approach? They add an id in the label and use that one

HenriqueLimas: Label added after this review

HenriqueLimas (approved):

PR #284, opened 2023-11-15
FEAT: EbayFakeMenu: support type/badgeNumber/badgeAriaLabel

darkwebdev:
image

### DIFF
--- a/src/ebay-fake-menu/menu-item.tsx
+++ b/src/ebay-fake-menu/menu-item.tsx
@@ -47,6 +47,7 @@ const EbayMenuItem: FC<EbayFakeMenuItemProps> = ({
     }
 
     const tick = <EbayIcon name="tick16" />
+    // todo: remove this workaround when Skin team fixes https://github.com/eBay/skin/issues/2208
     const badgeStyleFix = {
         marginLeft: 'var(--spacing-100)',
         marginRight: 'var(--spacing-100)'
--- a/src/ebay-fake-menu/menu-item.tsx
+++ b/src/ebay-fake-menu/menu-item.tsx
@@ -1,22 +1,34 @@
 import React, { ComponentProps, FC, useEffect, useRef } from 'react'
 import classNames from 'classnames'
 import { EbayIcon } from '../ebay-icon'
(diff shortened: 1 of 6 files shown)


src/ebay-fake-menu/menu-item.tsx
```diff
@@ -1,40 +1,81 @@
...
+            hasBadge && 'menu__item--badged'),
+        'aria-label': badgeAriaLabel
+    }
+
+    const tick = <EbayIcon name="tick16" />
+    const badgeStyleFix = {
+        marginLeft: 'var(--spacing-100)',
+        marginRight: 'var(--spacing-100)'
+    }
+    const badge = hasBadge && <EbayBadge type="menu" number={badgeNumber} style={badgeStyleFix} />
```
montoya332, line 54: LGTM but maybe we can create another helper/util component called EbayBadgeTick to avoid overloading this component with edge cases Fix

src/ebay-fake-menu/menu-item.tsx
```diff
@@ -1,40 +1,81 @@
...
             aria-disabled={disabled ? 'true' : undefined}
         >
             <span>
                 {children}
+                {badge}
             </span>
-            <EbayIcon name="tick16" />
+            {tick}
         </a>
     )
```
montoya332, line 80: Trying to think of ways we can avoid DRY and make it easier to read . probably a bit of overkill but what do you think 
```
const content = (
    <span>
        {children}
        {badge}
    </span>
);

const Component = type === 'button' ? 'button' : 'a';

const props = type === 'button'
    ? { ...itemProps, type: 'button', disabled }
    : { ...itemProps, 'aria-disabled': disabled ? 'true' : undefined };

return (
    <Component {...props}>
        {content}
        {tick}
    </Component>
);
```

darkwebdev: I've thought about it, but decided that hurts readability. DRY is good when it's not extreme :)

HenriqueLimas (approved):

montoya332 (approved):

PR #288, opened 2023-11-15
FEAT: stricter TS - step 1

darkwebdev:
- strict:true
- noImplicitThis: true
- strictBindCallApply: true
- useUnknownInCatchVariables: true


src/common/debounce.ts
```diff
@@ -1,10 +1,9 @@
...
-    return () => {
-        clearTimeout(timer)
-        timer = setTimeout((...args) => {
-            timer = null
-            fn.apply(this, ...args)
-        }, ms)
-    }
-}
+export const debounce =
+  <F extends (...args: Parameters<F>) => ReturnType<F>>(fn: F, ms: number): (...args: Parameters<F>) => void => {
```
HenriqueLimas, line 2: Should we use `function` instead of arrow function here? The `=> void => {}` confuses where is the return type definition ends and when is the arrow function starting

shpandian (approved):
LGTM

PR #290, opened 2023-11-16
FEAT: stricter TS - step 2

darkwebdev:
`strictFunctionTypes` = true


src/common/component-utils/utils.ts
```diff
@@ -1,23 +1,49 @@
...
+        .map(comp => typeof comp === 'function' && comp.name)
+        .filter(Boolean)
+    return filterBy(nodes, child => types.includes(elementType(child)))
 }
 
-export function filterBy(nodes: ReactNode = [], predicate: (el: ReactElement) => boolean): ReactElement[] {
-    const elements = Children.toArray(nodes) as ReactElement[]
-    return elements.filter(predicate)
+// Filters out nodes of certain component type(s)
+export const excludeComponent = (nodes: ReactNode = [], componentType: FC<any> | FC<any>[]): ReactNode[] => {
```
HenriqueLimas, line 43: Should we use `unknown` instead of `any`?

src/ebay-button/button.tsx
```diff
@@ -27,7 +28,7 @@ export type EbayButtonProps = {
     split?: Split;
     transparent?: boolean;
     onClick?: (e: MouseEvent) => void;
-    onEscape?: (e: KeyboardEvent<HTMLButtonElement | HTMLAnchorElement>) => void;
+    onEscape?: EbayKeyboardEventHandler<HTMLButtonElement & HTMLAnchorElement>;
```
HenriqueLimas, line 31: Isn't this a breaking change? Should we maintain `|` instead of `&`?

src/ebay-infotip/ebay-infotip.tsx
```diff
@@ -67,27 +48,35 @@ const EbayInfotip: FC<InfotipProps> = ({
...
+    const { children: contentChildren, ...contentProps } = elementProps(content)
+
+    const buttonProps = {
+        ref: buttonRef,
+        onClick: toggleTooltip,
+        disabled,
+        variant,
+        'aria-label': ariaLabel,
+        'aria-expanded': isExpanded,
+        icon
```
ronikadriu, line 60: Should we also install eslint for typescript, so  it can also give us error when we forget `;`

src/common/component-utils/utils.ts
```diff
@@ -1,23 +1,49 @@
...
+              return undefined
+      }
+  }
 
-export function excludeComponent(nodes: ReactNode = [], componentType: FC): ReactElement[] {
-    const elements = Children.toArray(nodes) as ReactElement[]
-    return elements.filter(({ type }) => type !== componentType)
-}
+// Returns props of a valid ReactElement
+export const elementProps = (node: ReactNode): Record<string, any> => isValidElement(node) ? node.props : {}
```
ronikadriu, line 19: I think the props of each component should be defined on component level, because what is the use of typrescript than when the second argument is any and one function for all components

src/ebay-infotip/ebay-infotip.tsx
darkwebdev: @ronikadriu we don't use `;` in this project historically :)

src/ebay-infotip/types.ts
```diff
@@ -1 +1,22 @@
...
+  variant?: Variant;
+  icon?: Icon;
+  disabled?: boolean;
+  initialExpanded?: boolean;
+  pointer?: PointerDirection;
+  overlayStyle?: CSSProperties;
+  onExpand?: () => void;
+  onCollapse?: () => void;
+  a11yCloseText: string;
+  'aria-label'?: string;
```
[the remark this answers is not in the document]
darkwebdev: no, this is the one that's passed on to HTML element. And for example `a11yCloseText` is a custom prop that doesn't exist in HTML.

Issue #296, opened 2023-12-06
EbaySwitch: Controlled prop not working


maxrchung:
The `EbaySwitch` component allows a `checked` prop to control its value, but on change this prop does not get enforced properly. Other controlled components like `EbayCheckbox` don't have this issue. I have a small sandbox and demo video to demonstrate issue: https://codesandbox.io/p/sandbox/dreamy-wozniak-xqlpqc?file=%2Fsrc%2FApp.tsx%3A10%2C32

https://github.com/eBay/ebayui-core-react/assets/3955187/63db2a94-dd9d-4cf3-be2a-b03debb1709d

I need this component to be controlled so I can set its state as a part of making an API call. Here is my flow:
1. Toggle switch on.
2. Make API call.
3. Disable switch while call is in progress.
4. Enable switch when call finishes.
5. Toggle switch on/off depending on API result.

With the current component I can't do step 5 as the component will always change the switch's states. I could do something kind of funny like throw an exception in my change handler but I that's probably not advised.

I will try and open a corresponding PR to fix this change.

darkwebdev:
@maxrchung thanks for the issue!

maxrchung:
> I've fixed your sandbox example: https://codesandbox.io/p/sandbox/billowing-forest-m55jr6?file=%2Fsrc%2FApp.tsx%3A24%2C20

Thanks, but maybe I am misunderstanding, how does this fix the sandbox? I was trying to show that the controlled property wasn't working.

PR #297, opened 2023-12-06
fix: added controlled logic for switch

maxrchung:
This PR addresses https://github.com/eBay/ebayui-core-react/issues/296. It doesn't seem like I have permissions to directly link issue.

This change allows `EbaySwitch` to be controlled by `checked` prop.

## Notes
- Added similar `isControlled` logic from `EbayCheckbox` into `EbaySwitch`.
- Added `defaultChecked` prop to follow logic from `EbayCheckbox`.

## Testing
- Added unit test to verify checkbox is not changed.
- Added story that changes `checked` prop.

## Demo
Showing issue before and updated stories after:

https://github.com/eBay/ebayui-core-react/assets/3955187/9d0658f3-1697-4d99-96c6-210b21b38c58

### DIFF
--- a/src/ebay-switch/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/src/ebay-switch/__tests__/__snapshots__/index.spec.tsx.snap
@@ -1,5 +1,35 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Storyshots ebay-switch Controlled switch-button 1`] = `
+<DocumentFragment>
+  <span
+    class="field"
+  >
+    <span
+      class="switch"
+    >
+      <input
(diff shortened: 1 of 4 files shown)


HenriqueLimas (changes requested):
Thank you for the PR. Usually we do this type of condition with `defaultChecked` and `checked` arguments. When `defaultChecked` is passed then it is an uncontrolled component, and when `checked` is passed that is controlled. We should do the same here for consistency. This will probably be a breaking change though. You can check how we do for ebay-checkbox [here](https://github.com/eBay/ebayui-core-react/blob/503d5f7e841999fcb9cdca1f24baf1f810426516/src/ebay-checkbox/checkbox.tsx#L26)

PR #299, opened 2023-12-15
Update docs: Add reference to floating labels CSS

bjoernr-de:
Updated documentation to include the necessary CSS for proper functioning of floating labels.

When utilizing the `ebay-textbox` element alongside floating labels, it didn't function as expected. The solution was to include the floating-label CSS file, which resolved the issue.

I presume the same applies to `ebay-listbox-button` and `ebay-select` elements, although I have **not** tested it.


HenriqueLimas (approved):
Thank you for adding this to our doc 🎉

PR #300, opened 2023-12-19
fix: dialog document attach event timeout

FiodorGherasimenco:
Issue related to react `18.2.0` witch will cause `baseDialog` to invoke `useEffect` when updating state for the content inside dialog.  It may happen that attach document click will happen after call cleanup inside useEffect.

Fix: clean timeout each time when useEffect cals cleanup.

### DIFF
--- a/src/ebay-dialog-base/components/dialogBase.tsx
+++ b/src/ebay-dialog-base/components/dialogBase.tsx
@@ -10,7 +10,8 @@ import React, {
     MouseEventHandler,
     ReactNode,
     KeyboardEvent,
-    KeyboardEventHandler
+    KeyboardEventHandler,
+    MouseEvent
 } from 'react'
 import classNames from 'classnames'
 import * as screenreaderTrap from 'makeup-screenreader-trap'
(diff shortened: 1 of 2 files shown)


HenriqueLimas (changes requested):
Thank you for the PR but we can’t do this change. This will break the current behavior on clicking outside of the page. As you can see the test were failing, we should not modify that test. We will need to have a different solution. Did you try to use event.stopPropagation() in your on background click?

FiodorGherasimenco:
> Thank you for the PR but we can’t do this change. This will break the current behavior on clicking outside of the page. As you can see the test were failing, we should not modify that test. We will need to have a different solution. Did you try to use event.stopPropagation() in your on background click?

`stopPropagation` will prevent click on close button so its also not a solution.

oxala (approved):

oxala:
Ops, sorry, wrong place for approval. Don't mind me. 🤐

src/ebay-dialog-base/components/dialogBase.tsx
```diff
@@ -96,6 +96,7 @@ export const DialogBase: FC<DialogBaseProps<HTMLElement>> = ({
     }, [])
 
     useEffect(() => {
+        let timeout: NodeJS.Timeout
```
HenriqueLimas, line 99: I think this should be a `number` instead since setTimeout inside an useEffect runs only on the browser

FiodorGherasimenco: So for IDE TypeScript there 2 types of `setTimeout` with/without prefix `window`. Without prefix setTimeout return type will be `NodeJS.Timeout` and for `window.setTimeout` it will be number.

Screenshot 2023-12-21 at 13 54 57

Screenshot 2023-12-21 at 13 50 21

I will change type to number but this will require `window` prefix.

HenriqueLimas (approved):

PR #311, opened 2023-12-29
Update with latest Skin 17 icons

HenriqueLimas:
#305


src/ebay-icon/types.ts
```diff
@@ -681,8 +649,12 @@ export type Icon =
     'unselectAll24' |
     'upload16' |
     'upload24' |
-    'verifiedCondition16Fit' |
-    'verifiedCondition24Fit' |
+    'venmo12Colored' |
```
shpandian, line 652: question outside of this - now that this has most of payment method icons do we still need our sprite images?

shpandian (approved):
LGTM

src/ebay-icon/types.ts
HenriqueLimas: Probably not, me and @montoya332 were discussing that maybe our payment sprites can use skin instead now

Issue #306, opened 2023-12-29
Use chevron 12 instead of 16 for dropdowns


HenriqueLimas:
Components to update:

- Breadcrumbs
- Field dropdown
- Select
- Fake menu button
- fake menu
- Listbox button
- Menu button
- Split button

PR #312, opened 2023-12-29
Update chevron down to 12 instead of 16

HenriqueLimas:
- Update button expand and listbox-button (other components uses a variation of these two components)

#306


shpandian (approved):
LGTM

PR #313, opened 2024-01-04
BREAKING Skin17

HenriqueLimas:
#311 
#312


shpandian:
@HenriqueLimas how about rest of the skin changes? are all they in?

shpandian (approved):
LGTM

HenriqueLimas:
@shpandian all the other changes defined in the [project](https://github.com/orgs/eBay/projects/10) are not required for the Breaking change, we can work as we go so we don't block SWU

PR #344, opened 2024-05-17
fix: ejs lacks certain pollution protection

imhunterand:
Affected of this project `EBay/ebayui-core-react` are vulnerable to Improper Control of Dynamically-Managed Code Resources due to the lack of certain pollution protection mechanisms. An attacker can exploit this vulnerability to manipulate object properties that should not be accessible or modifiable.

```js
function Template(text, opts) {
  opts = opts || utils.createNullProtoObjWherePossible();
  var options = utils.createNullProtoObjWherePossible();
  this.templateText = text;
  /** @type {string | null} */
	@@ -949,3 +949,4 @@ exports.name = _NAME;
if (typeof window != 'undefined') {
  window.ejs = exports;
}
```
CVE-2024-33883
CWE-1321

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5197,9 +5197,9 @@ ee-first@1.1.1:
   integrity sha512-WMwm9LhRUo+WUaRN+vRuETqG89IgZphVSNkdFgeb6sS/E4OrDIN7t48CAewSHXc6C8lefD8KKfr5vY61brQlow==
 
 ejs@^3.1.8:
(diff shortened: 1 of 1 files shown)


HenriqueLimas:
Thank you for the PR, we have a denpendabot PR that made the same changes here #342

HenriqueLimas:
Thank you for the PR, we have a denpendabot PR that made the same changes here #342

PR #352, opened 2024-07-10
feat(Icon): Update Icon class names

HenriqueLimas:
Update class name for icons from `icon--{name}` to `icon--size`.


darkwebdev (approved):
seems fine

PR #356, opened 2024-08-08
fix: removes unnecessary 'role' in <section ...> elements

rglombov:
According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/region_role#prefer_html, it is not necessary to specify role to `<section>` elements.

### DIFF
--- a/src/ebay-page-notice/page-notice.tsx
+++ b/src/ebay-page-notice/page-notice.tsx
@@ -47,8 +47,7 @@ const EbayPageNotice: FC<Props> = ({
         <section
             {...rest}
             aria-labelledby={id || `${status}-status`}
-            className={`page-notice ${status !== `general` ? `page-notice--${status}` : ``}`}
-            role="region">
+            className={`page-notice ${status !== `general` ? `page-notice--${status}` : ``}`}>
             {status !== `general` ? (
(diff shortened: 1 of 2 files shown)


abiyasa:
As discussed on Slack with eBay Skin and accessibility team, we will keep both `<section>` and `role-region` even though it's redundant:

- This is aligned with eBay Skin Page & Section notice (see https://opensource.ebay.com/skin/#page-notice or https://opensource.ebay.com/skin/#section-title), and also eBay UI Core component (e.g [ebay-section-notice](https://github.com/eBay/ebayui-core/blob/5659ebe36d1c41184284668887168c8a3920988f/src/components/ebay-section-notice/index.marko#L12)) which eBay UI Core React is based on. We should not diverge from the official components for consistency (see [eBay MIND pattern](https://ebay.gitbook.io/mindpatterns#pattern-principles))
- This redundant was because the Platform team weren't completely sure if `<section>` is supported by AT in all scenarios. So, having a fallback with `role=region` helps us achieve that. Based on this [table from a11ysupport.io](https://a11ysupport.io/tech/html/section_element#support-table-2), JAWS and Talkback still partially support `<section>` but this will change in the future

I am closing this ticket as won't fix/change.

PR #360, opened 2024-09-10
fix: fixed drawer windowClass prop

oxala:
Currently the `windowClass` prop is swallowed by the drawer component and not being passed to the DialogBase. This PR fixes that.


darkwebdev:
@oxala Hey Gleb! Thanks for the PR!
One question: any idea if that works the same way in eBayUI marko?

oxala:
It's also broken in marko: https://github.com/eBay/ebayui-core/blob/master/src/components/ebay-drawer-dialog/index.marko#L19-L23

I know that you trying to be a 1:1 replica of the marko library. But replicating a bug, since i CAN pass the windowClass, is not a good way in my opinion.

darkwebdev (approved):

PR #366, opened 2024-10-03
core: Use Vite for bundling components

HenriqueLimas:
- Use Vite instead of `tsc` for bundling the files
- Use one single tsconfig and upgrade module properties
- Use one single eslintrc config

This builds only CJS files since having both ESM and CJS might increase and import both files until all the libraries that uses ui-core-react also migrated to ESM. This is not a breaking change since it is still CJS only.

In some applications internally this reduced vendor file by 3kb (gzip)

Available to test with v7.1.0-alpha.3


shpandian (approved):
LGTM 

curious, if its only CJS(which is what tsc did as well right), vite bundling saves 3KB 🥇

HenriqueLimas:
@shpandian TSC doesn't do tree-shaking, it just compile everything, but vite does it

PR #367, opened 2024-10-07
support unique ids on ebay-section

woorea:
Generate the same id section-notice-information when there are multiple SectionNotices causing invalid HTML and  duplicate-id-aria on accessibility tools

### DIFF
--- a/src/ebay-section-notice/section-notice.tsx
+++ b/src/ebay-section-notice/section-notice.tsx
@@ -5,7 +5,7 @@ import React, {
     KeyboardEventHandler,
     MouseEvent,
     MouseEventHandler,
-    ReactElement, useState
+    ReactElement, useId, useState
 } from 'react'
 import cx from 'classnames'
 import { EbayNoticeContent } from '../ebay-notice-base/components/ebay-notice-content'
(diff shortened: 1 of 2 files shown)


src/ebay-section-notice/section-notice.tsx
```diff
@@ -5,7 +5,7 @@ import React, {
     KeyboardEventHandler,
     MouseEvent,
     MouseEventHandler,
-    ReactElement, useState
+    ReactElement, useId, useState
```
HenriqueLimas, line 8: `useId` is only available on React 18, and we still have applications that runs in React 16, this will be a huge breaking change. For now we can do the same approach used on [EbayIcon](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-icon/icon.tsx#L27), where we use a state and a useEffect to set on mount. Reason we do on mount is to make sure the server and client rendering matches.

HenriqueLimas (changes requested):
Thanks for the PR! That is great, just that we need to consider that some applications internally still uses React 16, so we need to make sure we don't break on their side

HenriqueLimas (approved):

PR #371, opened 2024-10-15
feat(EbaySvg): Support list of icons to render

HenriqueLimas:
- Add a new `icons` array that will render only the necessary icon by the application
- By default, renders all the icons so we don't introduce a breaking change
- Change the icon generation script to create the new `symbols.tsx` file


darkwebdev (approved):
awesome, thanks!

shpandian (approved):
finally, long pending :)  hopefully apps make use of it and let it be in default state.

@HenriqueLimas lets spread the word in the react channel post publish ..

PR #372, opened 2024-10-15
fix: TypeScript module resolution

HenriqueLimas:
# Description
Rename files to be `/<component-name>/index.js` instead of `/<component-name>.js` so it maintains typescript definition files

# Context

The current `dist` folder structure is:

```
/dist
    /ebay-button
          /index.d.ts
   ebay-button.js
```

This is causing an issue when imported internally (for example the `ebay-dialog-base`) where typescript is conflicting with which file to solve (is it `root/name/index.d.ts` or is it `root/name.js`).

### DIFF
--- a/vite.config.mjs
+++ b/vite.config.mjs
@@ -37,7 +37,7 @@ export default defineConfig({
     build: {
         lib: {
             entry: componentEntries,
-            fileName: "[name]",
+            fileName: "[name]/index",
             // Use CommonJS only until we upgrade all packages that uses ui-core-react to use ESM.
             // If we use ESM, the bundle might have both ESM and CJS, which will increase the bundle size,
             // and might cause reference issues.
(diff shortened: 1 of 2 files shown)


vite.config.mjs
```diff
@@ -37,7 +37,7 @@ export default defineConfig({
     build: {
         lib: {
             entry: componentEntries,
-            fileName: "[name]",
+            fileName: "[name]/index",
```
shpandian, line 40: wont it get confused with actual folder structure?

shpandian: i meant for the fact that its only reflecting on the build?

shpandian (approved):
LGTM - hope minor version is still fine.

vite.config.mjs
shpandian: checked offline. we are ok to proceed.

PR #373, opened 2024-10-18
fix(Icon): Add react import on icon symbols

HenriqueLimas:
This fixes `React` is not defined when using classic react jsx configurations


aaron-nance (approved):
Looks good. Thanks Henrique!

PR #375, opened 2024-11-02
fix(DialogBase): Don't trigger the hide animation on strict mode

HenriqueLimas:
## Description
- Change to use a previousOpen value to make sure we show the hide animation only when it changes.
- Add StrictMode to storybook

## Context

The previous logic was checking if it was the first render, but on StrictMode the layout effect is executed twice, triggering the hide animation logic during development.

### DIFF
--- a/src/ebay-dialog-base/components/animation.ts
+++ b/src/ebay-dialog-base/components/animation.ts
@@ -30,7 +30,7 @@ export function useDialogAnimation({
     enabled,
     onTransitionEnd
 }: DialogAnimationHookProps): void {
-    const firstRender = useRef(true)
+    const previousOpenValue = useRef(open)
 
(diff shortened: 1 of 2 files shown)


oxala:
Sweet, thanks!

src/ebay-dialog-base/components/animation.ts
```diff
@@ -54,7 +54,9 @@ export function useDialogAnimation({
                 classPrefix,
                 onTransitionEnd
             })
-        } else if (!firstRender.current) {
+        // Trigger the hide animation only when that "open" value changed to make sure it doesn't flicker the dialog.
```
shpandian, line 57: good find.. is there any other component would get into the similar logic?

shpandian (approved):
LGTM

PR #379, opened 2024-12-12
fix(EbaySvg): Hide SVG definition from screen readers

HenriqueLimas:
Currently screen readers are announcing the SVG icon definitions when hitting the page. This PR is to hide them from screen readers


shpandian (approved):
LGTM

PR #381, opened 2024-12-19
feat(EbayListbox): Add EbayListbox and use it with ListboxButton

HenriqueLimas:
## Description

- Added `EbayListbox` component
- Use makeup js packages for handling behavior and accessibility
- Follow logic for `EbayListboxButton` same as [marko implementation](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-listbox-button) for consistency and accessibility considerations
- Add missing features that was present in the marko version

Note, this is not a breaking change as the interface follows the exact same as before (as it can be seen in the storybook and tests)


HenriqueLimas:
There is one issue by using makeup-js:

- For performance reason we don't render the list on first render, we do on click instead. The makeup-js expander used on listbox button needs to have the content to expand available in order to add the event listeners. Closing this PR for now, we will consider a better approach.

PR #382, opened 2024-12-20
fix(EbayListboxButton): Show listbox with active descendent on button focus

HenriqueLimas:
- Show the `listbox` on focus of the button since on click the screen reader cannot check for it on focus
- Update `aria-activedescendent` logic to render function instead of manually setting the attributes of the ref
- Move `aria-activedescendent` to `listbox` component instead of div container


shpandian (approved):
LGTM

PR #384, opened 2024-12-23
FEAT: Ebay Toggle Button and Ebay Toggle Button Group

bruno2kd:
I can't create a new branch for this feature on the repo, so if you want me to close this and create a new PR on a new branch, please let me know.

Issue: https://github.com/eBay/ebayui-core-react/issues/281

I am working on collectibles and design asked for this component. I built for our app and I am creating a PR here with the code we have in case you want this contribution.

Heavily based on Marko's implementation:
https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-toggle-button-group
https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-toggle-button

If you want, I can reach me on slack in the handle bruduarte

thanks

### DIFF
--- a/src/ebay-toggle-button-group/toggle-button-group.tsx
+++ b/src/ebay-toggle-button-group/toggle-button-group.tsx
@@ -1,10 +1,7 @@
 import React, { ReactElement, cloneElement, useState, useCallback } from 'react'
 import cx from 'classnames'
 import { filterByType } from '../common/component-utils'
-import {
-    ToggleButtonEvent,
-    ToggleButtonProps
-} from '../ebay-toggle-button/types'
+import { ToggleButtonProps } from '../ebay-toggle-button/types'
 
 import { EbayToggleButton } from '../ebay-toggle-button'
 import { ToggleButtonGroupProps } from './types'
@@ -36,7 +33,7 @@ const EbayToggleButtonGroup: React.FC<ToggleButtonGroupProps> = ({
     >(getInitialPressedButtons)
 
     const handleToggle = useCallback(
-        (toggleEvent: ToggleButtonEvent, index: number) => {
+        (toggleEvent: React.MouseEvent<HTMLButtonElement>, index: number) => {
             setPressedButtons((prevState: Record<number, boolean>) => {
                 let newState: Record<number, boolean> = {}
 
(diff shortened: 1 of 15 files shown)


src/ebay-toggle-button-group/toggle-button-group.tsx
```diff
@@ -0,0 +1,101 @@
...
+                }
+                onChange(toggleEvent, newState)
+                return newState
+            })
+        },
+        [variant, onChange]
+    )
+
+    useEffect(() => {
+        setPressedButtons(
```
HenriqueLimas, line 62: Lets use the initial state instead of using a use effect. So instead of starting with `{}` we should start with this logic. `useEffect` should be avoided as much as possible and the initial state will help here:

src/ebay-toggle-button-group/toggle-button-group.tsx
```diff
@@ -0,0 +1,101 @@
...
+            data-columns-min={columnsMin}
+            data-columns-xs={columnsXS}
+            data-columns-sm={columnsSM}
+            data-columns-md={columnsMD}
+            data-columns-xl={columnsXL}
+            {...rest}
+        >
+            <ul aria-label={a11yText} aria-labelledby={a11yLabelId}>
+                {buttons.map((button: ReactElement, i) => {
+                    const { ...buttonRest }: ToggleButtonProps = button.props
```
HenriqueLimas, line 82: No need to pass `rest` this is already taken care by the `cloneElement`

src/ebay-toggle-button/toggle-button.tsx
```diff
@@ -0,0 +1,87 @@
...
+    title,
+    subtitle,
+    layoutType = 'minimal',
+    icon,
+    img,
+    onToggle,
+    ...rest
+}) => {
+    const handleClick = (e: React.MouseEvent) => {
+        onToggle({ originalEvent: e, pressed: !pressed })
```
HenriqueLimas, line 19: On React we use the react (and dom) way of sending events. So it will be `onToggle(event, data)`. We have a helper type that you can use https://github.com/eBay/ebayui-core-react/blob/main/src/common/event-utils/types.ts

HenriqueLimas (changes requested):
Thanks for the PR! Left some comments to be aligned with the React API and improve some testing

HenriqueLimas (approved):

PR #395, opened 2025-01-09
core: Add changesets releases

HenriqueLimas:
- Add storybook deployment to GH pages on `push` and `pr` actions
- Setup [changesets](https://github.com/changesets/changesets) to the repo
- Add github action for release using changesets


.changeset/honest-elephants-care.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ebay/ui-core-react": patch
```
agliga, line 2: Seems more like a feature, so Id put it in a minor version

.github/actions/release/action.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+        - name: Enter prerelease mode
+          shell: sh
+          if: ${{ github.ref != 'refs/heads/main' }}
+          run: yarn prerelease:start
+
+        - name: Exit prerelease mode
+          shell: sh
+          if: ${{ github.ref == 'refs/heads/main' }}
+          run: yarn prerelease:end
```
agliga, line 16: While we do have a prerelase mode, we havent used it in recent times. Im considering to remove it since patches are very easy to put out.

package.json
```diff
@@ -100,9 +102,11 @@
...
     "copy": "copyfiles package.json README.md dist; yarn copy:rest",
     "copy:rest": "yarn copyfiles -u 1 'src/**/*.md' dist",
-    "release": "yarn clean && yarn build && yarn copy && CI=1 npm publish dist/ --access public",
-    "release:rc": "yarn clean && yarn build && yarn copy && CI=1 npm publish --tag rc dist/",
-    "prepare": "node scripts/husky-install || node ../scripts/husky-install"
+    "release": "yarn clean && yarn build && yarn copy && CI=1 changeset publish",
+    "prerelease:end": "changeset pre exit || :",
+    "prerelease:start": "changeset pre enter next || :",
+    "prepare": "node scripts/husky-install || node ../scripts/husky-install",
+    "version": "yarn update-icons && git add -A src && changeset version"
```
agliga, line 109: One issue we found was that when doing a release, the package-lock version does not get changed. By doing npm i --package-lock-only this will regenerate the package-lock version. That said, I don't know if theres an equivalent or similar problem for yarn.

.github/workflows/push.yml
```diff
@@ -38,3 +38,26 @@ jobs:
...
+    release:
+        needs: deploy
+        name: Release
+        # Sets permissions of the GITHUB_TOKEN to allow opening release PR
+        permissions:
+            id-token: write
+            contents: write
+            packages: write
+            pull-requests: write
+            issues: read
```
agliga, line 58: I believe these are need for the the changesets to work. I know at least pull-requests or packages is needed and Im not sure these will be used by the deploy action

agliga (approved):
Overall looks good, just a couple of minor comments. 
I think this just needs to be tested properly (and it might take a couple of tweaks)

.changeset/honest-elephants-care.md
HenriqueLimas: Technically I don't need a version for this, it was just to test the action

.github/actions/release/action.yml
HenriqueLimas: I though this one was for testing that the command would work or not before releasing it. Misunderstood the command. I will remove these scripts

package.json
HenriqueLimas: Does changeset changes the version of the lock file?

.github/workflows/push.yml
HenriqueLimas: These are what people are using it seems https://github.com/changesets/action/issues/179#issuecomment-1139116421

.changeset/honest-elephants-care.md
HenriqueLimas: I will remove this changeset. It is a `core` change that doesn't need a version update

Issue #394, opened 2025-01-08
Remove deprecated APIs


HenriqueLimas:
Remove code and documentation of:

- [x] Drawer dialog [onShow](https://github.com/eBay/ebayui-core-react/blob/9f2bece932207fee479593d370e4753938b7c379/src/ebay-drawer-dialog/components/drawer.tsx#L18)
- [x] EbayTabs [onTabSelect](https://github.com/eBay/ebayui-core-react/blob/9f2bece932207fee479593d370e4753938b7c379/src/ebay-tabs/types.ts#L18)
- [x] EbayTabs `index`[ property](https://github.com/eBay/ebayui-core-react/blob/9f2bece932207fee479593d370e4753938b7c379/src/ebay-tabs/README.md?plain=1#L31) 
- [x] Deprecated [skin 12 color](https://github.com/eBay/ebayui-core-react/blob/9f2bece932207fee479593d370e4753938b7c379/src/common/floating-label-utils/hooks.tsx#L33)
- [x] EbayButton [default size](https://github.com/eBay/ebayui-core-react/blob/9f2bece932207fee479593d370e4753938b7c379/src/ebay-button/types.ts#L5)

HenriqueLimas:
#397

PR #397, opened 2025-01-10
Remove deprecated apis

HenriqueLimas:
- [Remove onShow deprecated method type and documentation](https://github.com/eBay/ebayui-core-react/commit/7cc192899f83b10e9df0927926c83c2549d62419)
- [feat(EbayTabs): Remove depreacated onTabSelect](https://github.com/eBay/ebayui-core-react/commit/98f45595c11668dfe2ff1ff5e95ce330f60bbf27)
- [feat(EbayTabs): Use selectedIndex instead of index](https://github.com/eBay/ebayui-core-react/commit/0512cbf606e2cc153fccc0e429d17eb3773c966f)
- [feat(FloatingLabel): Remove unused old skin colors for autocomplete)](https://github.com/eBay/ebayui-core-react/commit/622cc55fe8127132aefb96db3a6786ed7faf69c9)
- [feat(EbayButton): Remove 'default' deprecated size option](https://github.com/eBay/ebayui-core-react/commit/c4236b222d6998386adabb32299bee7407c6d999)

Note: waiting for changesets PR to generate the changeset 

Fixes: #394


src/ebay-button/button.tsx
```diff
@@ -66,8 +66,7 @@ const EbayButton:FC<Props> = ({
     const sizeStyles: { [key in Size]: string } = {
         large: `${classPrefix}--large`,
         small: `${classPrefix}--small`,
-        regular: '',
-        default: ''
+        regular: ''
```
LuLaValva, line 69: Is there a reason to keep `regular` even though it has no styles? Maybe `Size` should just be `'small' | 'large' | undefined`

LuLaValva (approved):
Looks good to me, just one question (probably applies to Marko eBayUI as well)

src/ebay-button/button.tsx
HenriqueLimas: That is true, it is not necessary since skin doesn't use. @agliga should we remove on Marko for v15 also?

HenriqueLimas: Actually marko, doesnt' have it anymore. Let me remove it

agliga (approved):

PR #399, opened 2025-01-14
feat(EbayDetails): Add new EbayDetails component

HenriqueLimas:
# Description

- Add new `EbayDetails` component, following [ebayui-core implementation](https://github.com/eBay/ebayui-core/tree/accordion/src/components/ebay-details)
- Upgrade user-event to latest variable so it triggers the `toggle` event on clicking of the details
- Updated tests to use the new user-event version that uses async

The EbayDetails is required for the new EbayAccordion component

Note: waiting for #395 to add the changeset


src/ebay-details/README.md
```diff
@@ -0,0 +1,51 @@
+# EbayButton
```
darkwebdev, line 1: update component name here :)

darkwebdev (changes requested):

src/ebay-details/README.md
HenriqueLimas: Ops 🤦

PR #401, opened 2025-01-16
fix menu-button and svg key error and add storybook console logging

ericbiewener:
`<EbayMenuButton>` and `<EbaySvg>` have React key errors. This fixes those, and adds the `@storybook/addon-console` package so that you can see these errors during development on this codebase.


src/ebay-menu-button/menu-button.tsx
```diff
@@ -162,8 +162,7 @@ function labelWithPrefixAndIcon({ text, prefixId, prefixLabel, menuButtonLabel,
     ]
     const labelWithPrefix = [prefixLabelElement, menuButtonLabel || textLabelElement]
     const labelArray = [icon, labelWithPrefix].flat().filter(Boolean) as JSX.Element[]
-
-    return labelArray.length ? labelArray : null
```
ericbiewener, line 166: Seems fine to return an empty array rather than `null`.

src/ebay-menu-button/menu-button.tsx
```diff
@@ -162,8 +162,7 @@ function labelWithPrefixAndIcon({ text, prefixId, prefixLabel, menuButtonLabel,
...
+    return labelArray.map((item, i) => cloneElement(item, { ...item.props, key: i }))
```
ericbiewener, line 165: This change is breaking some tests. Not sure if it is related to storybook's jest integration, or something I'm doing wrong with `cloneElement`.

src/ebay-svg/symbols.tsx
HenriqueLimas, line 1: The SVG symbols is auto generated, we need to change the [script](https://github.com/eBay/ebayui-core-react/blob/main/scripts/update-icons-helpers.js) that update the icons instead

src/ebay-menu-button/menu-button.tsx
HenriqueLimas: It might be related to the Fragment maybe. Also, you don't need `...item.props` as [cloneElement](https://react.dev/reference/react/cloneElement#parameters already merge the props.

> for every prop in the props object, the returned element will “prefer” the value from props over the value from element.props. The rest of the props will be filled from the original element.props.

But for this use case, I think we can manually add the key for each element that is being created here. Something like:

```jsx
const textLabelElement = text.length ? <span key="text-label">{text}</span> : null
const prefixLabelElement = !prefixId && prefixLabel && [
  <span key="prefix-label" className="menu-button-prefix-label">{prefixLabel}</span>,
  <Fragment key="extra-space">&nbsp;</Fragment>
]
```

.storybook/preview.jsx
```diff
@@ -1,5 +1,6 @@
 import React, {StrictMode} from 'react'
 import { EbaySvg } from '../src/ebay-svg'
+import '@storybook/addon-console';
```
HenriqueLimas, line 3: Lets not add this one for now. We use `actions` tabs for showing the callbacks and this plugin also uses it, having this might be confusing. (we can miss let log anyway if you don't open this tab)

HenriqueLimas (changes requested):

.storybook/preview.jsx
ericbiewener: > we can miss the log anyway if you don't open this tab

Sure, but having your dev console open is a pretty standard part of development. Is there a way to see the error logs without this?

Regardless, removing per your request.

src/ebay-menu-button/menu-button.tsx
ericbiewener: > you don't need ...item.props

Just copied that pattern from [here](https://github.com/ericbiewener/ebayui-core-react/blob/39ea9b21515e3e3ca54b3373ddaa2768888bbe33/src/ebay-menu-button/menu-button.tsx#L146).

The resulting array can include [icon] and [menuButtonLabel](https://github.com/ericbiewener/ebayui-core-react/blob/39ea9b21515e3e3ca54b3373ddaa2768888bbe33/src/ebay-menu-button/menu-button.tsx#L163), both of which are provided by the consumer and thus aren't guaranteed to have a key. So those would at least need to be cloned.

No idea why the tests were failing, but making `prefixLabelElement` a single element rather than an array by wrapping the `<span>` in the fragment appears to have fixed it.

.storybook/preview.jsx
HenriqueLimas: It shows in the normal console on local. It doesn't show in the deployed version because it is a prod build, so warnings are removed by React

HenriqueLimas (approved):

PR #403, opened 2025-01-17
feat: Add support for floating-ui on dropdowns

HenriqueLimas:
# Description

Update dropdown components to support [floating-ui](https://floating-ui.com/) as per next skin version guidelines. These are the components that were updated:

- date-textbox
- fake-menu-button
- listbox-button
- menu-button

Note: other components like combobox or filter-menu are not current existing on react core

# Context

Used [`useFloating()`](https://floating-ui.com/docs/usefloating) hook and encapsulated with a `useFloatingDropdown` hook adding same behaviors as [marko implementation](https://github.com/eBay/ebayui-core/pull/2381/files#diff-eba6f0652c13f68755c5d59db5abf12ead1abd71ad23d745105f17c68b0b829cR36). 

The `useFloating()` can be used in different ways, like passing the reference as argument or receiving methods to set the reference. We used the latest approach so it is properly set when we render the overlay only when opened or focused.


HenriqueLimas:
@darkwebdev the issue is fixed in the next skin version, although not yet available. This is my local pointing to the new skin version:

Screenshot 2025-01-24 at 1 53 07 PM

agliga (approved):

package.json
```diff
@@ -72,6 +72,7 @@
     "@babel/register": "^7.0.0",
     "@changesets/cli": "^2.27.11",
     "@ebay/skin": "^18.4.0",
+    "@floating-ui/react": "^0.27.3",
```
shpandian, line 75: @HenriqueLimas why on dev dependencies? you need it as part of the component right?

HenriqueLimas: It should be dependency, thanks let me update

shpandian (approved):
LGTM

PR #412, opened 2025-01-22
feat(EbayListbox): Add new component

HenriqueLimas:
- Add new `EbayListbox` component following [marko implementation](https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-listbox) and API

Note: Using it on EbayListboxButton will be handle in a different PR as some changes are needed on makeup-js

Storybook link: https://opensource.ebay.com/ebayui-core-react/ebay-listbox/index.html?path=/docs/building-blocks-ebay-listbox--docs

### DIFF
--- a/src/ebay-listbox/README.md
+++ b/src/ebay-listbox/README.md
@@ -1,0 +1,100 @@
+# EbayListbox
+
+## Demo
+
+[Storybook](https://opensource.ebay.com/ebayui-core-react/main/?path=/docs/building-blocks-ebay-listbox--docs)
+
+## Import JS
+
+```jsx harmony
+import {
+    EbayListbox,
+    EbayListboxOption,
+    EbayListboxOptionDescription,
+} from "@ebay/ui-core-react/ebay-listbox";
+```
+
+## Import following styles from SKIN
+
+```js
+import "@ebay/skin/listbox";
+```
+
+## Import styles using SCSS/CSS
+
+```js
+import "@ebay/skin/listbox.css";
+```
+
+## Usage
+
+```jsx
+<EbayListbox>
+    <EbayListboxOption value="AA" text="Option 1" />
+    <EbayListboxOption value="BB" text="Option 2" />
+    <EbayListboxOption value="CC" text="Option 3" />
+</EbayListbox>
+```
+
+## Attributes
+
+### EbayListbox
+
+| Name                     | Type    | Required | Description                                                                                             |
+| ------------------------ | ------- | -------- | ------------------------------------------------------------------------------------------------------- |
+| `name`                   | string  | No       | Used for the name attribute of the native `<select>`.                                                   |
+| `disabled`               | boolean | No       | Set to true if the field is disabled.                                                                   |
+| `listSelection`          | string  | No       | If `manual`, the user will need to press enter to select an item using the keyboard. Otherwise, `auto` will automatically select as the user presses up/down. |
+| `typeaheadTimeoutLength` | number  | No       | Timeout length to pass to typeahead.                                                                    |
+| `maxHeight`              | string  | No       | Example: 100px, 200px, 10rem.                                                                           |
+
+### EbayListboxOption
+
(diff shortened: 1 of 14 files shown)


darkwebdev:
looks fine, the only thing that I'd love to see is either storybook controls setup for experimenting with various options or just more stories.

HenriqueLimas:
@darkwebdev thanks for the comment, added the arguments as story controls and also fixed a bug on the typeahead

src/ebay-listbox/README.md
```diff
@@ -0,0 +1,100 @@
+# EbayListbox
+
+## Demo
+
+[Storybook](https://opensource.ebay.com/ebayui-core-react/main/?path=/docs/building-blocks-ebay-listbox--docs)
```
shpandian, line 5: @HenriqueLimas there is an error in the URL path..the above link is broken.. PR desc has right link - https://opensource.ebay.com/ebayui-core-react/ebay-listbox/index.html?path=/docs/building-blocks-ebay-listbox--docs

src/ebay-listbox/listbox-option.tsx
```diff
@@ -0,0 +1,51 @@
...
+    selected,
+    ...rest
+}) => {
+    const description = filterByType(children, EbayListboxOptionDescription)
+    const displayText = text || (!description?.length ? children : '')
+
+    return (
+        <div
+            {...rest}
+            tabIndex={disabled ? -1 : tabIndex}
```
shpandian, line 31: is it ok to skip for a11y as well when its disabled?

src/ebay-listbox/README.md
HenriqueLimas: Yeah, it is because is not available in `main` branch yet, it will when we merge

src/ebay-listbox/listbox-option.tsx
HenriqueLimas: Yes so it cannot be accessed with the tab index

src/ebay-listbox/listbox.tsx
```diff
@@ -0,0 +1,217 @@
...
+                    })
+                )}
+            </div>
+
+            <select
+                hidden
+                className={classNames('listbox__native', selectClassName)}
+                name={name}
+                disabled={disabled}
+                value={options[selectedIndex]?.props.value}
```
agliga, line 206: I believe we use `selected` in `<option>` here. Part of the reason is for form submission. When using value it doesn't seem to be working properly
Here's a test I made: 
https://jsfiddle.net/pfoe3zhn/

HenriqueLimas: Make sense.  Let me update

LuLaValva: @agliga this is true in HTML but not React. React has special handling for `<select value>`. See docs [here](https://react.dev/reference/react-dom/components/select#controlling-a-select-box-with-a-state-variable)

LuLaValva: In fact, based on the docs `selected` is not supported at all in React

HenriqueLimas: React throws a warning to use `value` or `defaultValue` instead of selected

src/ebay-listbox/README.md
shpandian: my bad -- ebay-listbox is the branch name here, not the comp :)

src/ebay-listbox/listbox.tsx
```diff
@@ -0,0 +1,217 @@
...
+                break
+            case 'Space':
+            case 'Enter':
+                handleChange(event, activeDescendantRef.current.index, false)
+                break
+            default:
+                break
+        }
+
+        const itemIndex = typeaheadRef.current.getIndex(
```
shpandian, line 95: should we do null check after the ref exists to avoid errors -- or is it safe 100%?

shpandian (approved):
LGTM

Issue #417, opened 2025-01-29
EbayCarousel: Add auto play behavior


HenriqueLimas:
Add new `autoplay`, `autoplayInterval`, `paused` properties
Add `onPlay` and `onPause` callbacks

PR #438, opened 2025-02-07
feat(EbayCarousel): add support for autoplay behavior

HenriqueLimas:
# Description

- Add `autoplay` behavior on carousel with animation to loop the slides.
- The `play` is enabled by default, unless the reduced motion is enable in the device
- Controls disabling is ignored when the autoplay is enable since we loop the slides

## How the slide looping works

When detecting that it is the last slide,

1. Disable the transition
2. Move the last slide position before the first slide location
3. Move the focus of the carousel to the new position of the last slide
4. Enable the transition
5. Move the focus to the first slide
6. Remove the temporary position of the last slide

Storybook: https://opensource.ebay.com/ebayui-core-react/carousel-autoplay/?path=/story/navigation-disclosure-ebay-carousel--autoplay

#417

### DIFF
--- a/.changeset/chilly-fireants-learn.md
+++ b/.changeset/chilly-fireants-learn.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ui-core-react": minor
+---
+
+feat(EbayCarousel): add support for autoplay behavior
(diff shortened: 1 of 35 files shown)


darkwebdev (approved):

agliga (changes requested):
One important aspect which is needed is to disable autoplay when mouse is over the carousel. I didnt see this code but i might have missed it
In coreui marko we use interacting variable for this. Or something like that

HenriqueLimas:
Thanks @agliga, I will add that behavior

agliga (approved):

PR #440, opened 2025-02-07
fix(EbayFakeMenuButton): use btn-cell to add margin between chevron and action content

HenriqueLimas:
`FakeMenuButton` and `MenuButton` have a different HTML structure, here we are putting the `btn-cell` back as it was before  for fake menu button

Screenshot 2025-02-07 at 8 42 41 AM

Screenshot 2025-02-07 at 8 42 54 AM

https://opensource.ebay.com/ebayui-core-react/fix-fake-menu-button/index.html?path=/story/buttons-ebay-fake-menu-button--default

### DIFF
--- a/.changeset/honest-parents-remember.md
+++ b/.changeset/honest-parents-remember.md
@@ -1,0 +1,5 @@
+---
+"@ebay/ui-core-react": patch
+---
+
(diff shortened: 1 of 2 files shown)


aaron-nance (approved):
Changes look good.

PR #442, opened 2025-02-11
feat(EbayTristateCheckbox): Added component for tristate checkbox based on marko version

anwayar:
Similar component to marko version: https://github.com/eBay/ebayui-core/tree/master/src/components/ebay-tri-state-checkbox


src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    const iconChecked = size === 'large' ?
+        <EbayIcon name="checkboxChecked24" className="checkbox__checked" /> :
+        <EbayIcon name="checkboxChecked18" className="checkbox__checked" />
+    const iconUnChecked = size === 'large' ?
+        <EbayIcon name="checkboxUnchecked24" className="checkbox__unchecked" /> :
+        <EbayIcon name="checkboxUnchecked18" className="checkbox__unchecked" />
+    const iconMixed = size === 'large' ?
+        <EbayIcon name="checkboxMixed24" /> :
+        <EbayIcon name="checkboxMixed18" />
```
agliga, line 69: I would have all these inside an single if statement, rather than checking each time. 
In the marko, we have it like that because of the way we render it in the template, but if setting a variable, I would do:
```
let iconChecked = <EbayIcon name="checked">;
let iconUnchecked = <EbayIcon name="unchecked">
let iconMixed = <EbayIcon name="mixed">

if (size === "large") {
  iconChecked = <EbayIcon name="checked-large">;
  iconUnchecked = <EbayIcon name="unchecked-large">
  iconMixed = <EbayIcon name="mixed-large">
}
```
Again, above is mainly using pseudo classes :)

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
```diff
@@ -0,0 +1,111 @@
...
+            return iconMixed
+        }
+        return iconUnChecked
+    }
+    return (
+        <>
+            <span className={containerClass} style={{ ...style, alignItems: 'center' }}>
+                <input
+                    {...rest}
+                    aria-checked={isChecked}
```
agliga, line 85: I would rename this, `isChecked` seems more like a boolean, where as this can be true/false/mixed

agliga (commented):
Overall, looks good from a code perspective.
Just a couple of minor cleanups.
After those, Ill approve, but would like @HenriqueLimas to take a look to make sure this is good from a React perspective

anwayar:
> Overall, looks good from a code perspective. Just a couple of minor cleanups. After those, Ill approve, but would like @HenriqueLimas to take a look to make sure this is good from a React perspective

Might need to add couple of more stories too to test disabled state and mixed implementation

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
anwayar: Added @agliga

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
anwayar: Renamed the variable @agliga

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { EbayIcon } from '../ebay-icon'
+import { EbayLabel, EbayLabelProps } from '../ebay-field'
+import { findComponent } from '../common/component-utils'
+import { EbayChangeEventHandler, EbayFocusEventHandler, EbayKeyboardEventHandler } from '../common/event-utils/types'
+
+type Size = 'default' | 'large'
+type CheckboxState = 'true' | 'false' | 'mixed'
+type InputProps = Omit<ComponentProps<'input'>, 'size' | 'onChange' | 'onFocus' | 'onKeyDown'>
+type EbayTristateCheckboxProps = {
+    checked?: CheckboxState,
```
HenriqueLimas, line 12: We need to support the component as a controlled and as uncotrolled. Controlled means that the application maintains the state of the checkbox (by using the `checked` property) and our component will render based on the `checked` property.
Uncontrolled means that we control the state of the input, and that is defined by the `defaultChecked` property that define the initial state (argument passed on `useState()`) and we update accordingly. Marko combine both, but in react we have this control vs uncontrol [pattern](https://react.dev/reference/react-dom/components/input#controlling-an-input-with-a-state-variable)

HenriqueLimas (changes requested):
Thank you for the PR! Added some comments to be better aligned with latest standards and remove some of the workarounds that we currently have on Checkbox

src/ebay-tri-state-checkbox/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { EbayLabel } from '../../ebay-field'
+import { EbayTriStateCheckbox } from '../index'
+
+const meta: Meta<typeof EbayTriStateCheckbox> = {
+    component: EbayTriStateCheckbox,
+    title: 'form input/ebay-tri-state-checkbox',
+    argTypes: {
+        checked: {
+            options: ["false", "mixed", "true"],
+            type: { category: "Options" },
```
HenriqueLimas, line 13: It should be:

```
control: { type: 'select'}
```

Make sure that you are able to make changes in the storybook controls section

src/ebay-tri-state-checkbox/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+            description:
+                'Either "large" or "default". Sets the checkbox icon. Default is regular. For mweb this should be set to large. (Note: The dimensions of the checkbox will not change, but only the icon)',
+            table: {
+                defaultValue: {
+                    summary: "default",
+                },
+            },
+        },
+        onChange: {
+            action: "on-change",
```
HenriqueLimas, line 40: It is `onChange` instead of the marko `on-change`, same on line 50

src/ebay-tri-state-checkbox/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+            description: "Triggered on change",
+            table: {
+                category: "Events",
+                defaultValue: {
+                    summary: "{ originalEvent, value, checked }",
```
HenriqueLimas, line 45: Its `event, { value, checked }`, same on line 55

src/ebay-tri-state-checkbox/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+}
+
+export default meta
+
+export const Default: StoryFn<typeof EbayTriStateCheckbox> = (args) => (
+    <>
+        <p style={{ alignItems : "center" }}>
+            <EbayTriStateCheckbox
+                {...args}
```
HenriqueLimas, line 68: since we are passing the `...args` we don't need to set any property as it is already handle by the `argTypes` definition.

Consider this for all the stories bellow also

src/ebay-tri-state-checkbox/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+                value="123"
+                checked="mixed"
+                id="checkbox-11"
+                onChange={(e, props) => action('onChange')(e, props)}
+                onFocus={(e, props) => action('onFocus')(e, props)}
+                onKeyDown={(e, props) => action('onKeyDown')(e, props)}
+            />
+        </p>
+        <p style={{ alignItems : "center" }}>
+            <EbayTriStateCheckbox {...args} value="123" id="checkbox-12" size="large" />
```
HenriqueLimas, line 78: We don't need this second example, the `size` is handled by the `args`

src/ebay-tri-state-checkbox/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,118 @@
...
+        </p>
+    </>
+)
+
+export const WithLabel: StoryFn<typeof EbayTriStateCheckbox> = (args) => (
+    <>
+    <p style={{ alignItems : "center" }}>
```
HenriqueLimas, line 85: Lets use the `EbayField` instead of the `p` as it is how it should be used with label

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
```diff
@@ -0,0 +1,98 @@
...
+    defaultChecked = 'false' as CheckboxState,
+    skipMixed = false,
+    onChange = () => {},
+    onFocus = () => {},
+    onKeyDown = () => {},
+    children,
+    inputRef,
+    ...rest
+}) => {
+    const [checkboxState, setCheckboxState] = useState<CheckboxState>(checked || defaultChecked)
```
HenriqueLimas, line 35: We use only `defaultChecked` as the initial state. We don't use the state when it is controlled. If we want the to always use the state, we would need to add an `useEffect` to update the state as the `checked` value changes but it is not ideal (as mentioned in this react team [post](https://react.dev/learn/you-might-not-need-an-effect))

You can check the `EbayCheckbox` for reference as this is also handle [here](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-checkbox/checkbox.tsx#L34)

src/ebay-tri-state-checkbox/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,36 @@
...
+    })
+    describe('on checkbox-button key down', () => {
+        it('should fire an event', () => {
+            const spy = jest.fn()
+            render(<EbayTriStateCheckbox aria-label="checkbox" value="123" onKeyDown={spy} />)
+            const input = getByRole('checkbox')
+            fireEvent.keyDown(input)
+            expect(spy).toHaveBeenCalledWith(eventOfType('keydown'), { value: '123', checked: "false" })
+        })
+    })
```
HenriqueLimas, line 35: We need to also add tests that the right checkbox icon is rendered depending on the state.

HenriqueLimas (changes requested):

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+    let iconChecked = <EbayIcon name="checkboxChecked18" className="checkbox__checked" />
+    let iconUnchecked = <EbayIcon name="checkboxUnchecked18" className="checkbox__unchecked" />
+    let iconMixed = <EbayIcon name="checkboxMixed18" />
+    if (size === 'large') {
+        iconChecked = <EbayIcon name="checkboxChecked24" className="checkbox__checked" />
+        iconUnchecked = <EbayIcon name="checkboxUnchecked24" className="checkbox__unchecked" />
+        iconMixed = <EbayIcon name="checkboxMixed24" />
+    }
+
```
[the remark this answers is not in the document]
anwayar: sorry I forgot to update this, I was testing if it was working and forgot to push

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
```diff
@@ -0,0 +1,99 @@
...
+    children,
+    inputRef,
+    ...rest
+}) => {
+    const [checkboxState, setCheckboxState] = useState<CheckboxState>(defaultChecked)
+    const currentCheckboxState = isControlled(checked) ? checked : checkboxState
+    const handleChange = (e: ChangeEvent<HTMLInputElement>) => {
+        const input = e.target
+        let newCheckboxState: CheckboxState = checkboxState
+        if (checkboxState === 'true') {
```
HenriqueLimas, line 41: we need to use `currentCheckboxState` in these if conditions

src/ebay-tri-state-checkbox/tri-state-checkbox.tsx
```diff
@@ -0,0 +1,99 @@
...
+        } else if (checkboxState === 'false' && !skipMixed) {
+            newCheckboxState = 'mixed'
+        } else {
+            newCheckboxState = 'true'
+        }
+        onChange(e, { value: input?.value, checked: newCheckboxState })
+        setCheckboxState(newCheckboxState)
+    }
+    const handleFocus = (e: FocusEvent<HTMLInputElement>) =>
+        onFocus(e, { value: e.target?.value, checked: checkboxState })
```
HenriqueLimas, line 52: Also here we should use `currentCheckboxState`

HenriqueLimas (approved):
Thank you so much for the PR 🎉

PR #443, opened 2025-02-11
fix: use target es2015 instead of esnext for better compatibility

HenriqueLimas:
nullish coalescing (`??`) is not supported on Webpack 4, updating the typescript target so it supported accross different builders.


aaron-nance (commented):
While this is a minor change, aren't all our eBay applications supposed to be on Webpack 5 by now?

aaron-nance (approved):
As we still need this, approved. :-)

PR #459, opened 2025-03-11
fix(Select): remove floating-label--inline class on first render

HenriqueLimas:
# Description

- Remove `floating-label__label--inline` from our side as it is handle by makeup-floating-label


agliga (approved):
Lgtm
I personally i would have the package nodejs 16 change in a separate commit but I'm fine with it as is.

PR #465, opened 2025-03-15
feat: Add EbayFilterMenuButton component

HenriqueLimas:
- Update FilterMenu to properly sets the tabindex
- Allow passing ref to FilterMenu
- Add new FilterMenuButton

Storybook: https://opensource.ebay.com/ebayui-core-react/filter-menu-button/?path=/docs/buttons-ebay-filter-menu-button--docs

### DIFF
--- a/src/ebay-filter-menu/filter-menu.tsx
+++ b/src/ebay-filter-menu/filter-menu.tsx
@@ -1,5 +1,17 @@
-import React, { ComponentProps, FC, Fragment, MouseEvent, useState } from 'react'
-import { filterByType, findComponent } from '../utils'
+import React, {
+    ComponentProps,
+    FC,
+    Fragment,
+    KeyboardEvent,
+    MouseEvent,
+    RefObject,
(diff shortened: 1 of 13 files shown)


src/ebay-filter-menu/filter-menu.tsx
```diff
@@ -208,4 +255,4 @@ const EbayFilterMenu: FC<EbayFilterMenuProps> = ({
     )
 }
 
-export default EbayFilterMenu
+export default withForwardRef(EbayFilterMenu)
```
shpandian, line 258: are we not on react 19? should we just use this as a PROP?

HenriqueLimas: No [we do support](https://github.com/eBay/ebayui-core-react/blob/main/package.json#L57-L58) React 16, we will when we upgrade the apps and will release a breaking change

shpandian (approved):
LGTM

Issue #433, opened 2025-02-05
Add EbayFilePreviewCard


HenriqueLimas:
https://opensource.ebay.com/ebayui-core/?path=/docs/media-ebay-file-preview-card--documentation

PR #475, opened 2025-04-07
#433: Add EbayFilePreviewCard

bruno2kd:
Issue: [#433](https://github.com/eBay/ebayui-core-react/issues/433)

## Demo:

https://github.com/user-attachments/assets/c77512d2-6476-41e0-bc3d-a32b27d6e90f

## Coverage:

Screenshot 2025-04-06 at 21 34 52


HenriqueLimas (changes requested):
Thanks for the PR. Added some comments to make sure we have same logic as the marko version

src/ebay-file-preview-card/file-preview-card.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    onClick={onSeeMore}
+                    aria-label={a11ySeeMoreText}
+                >
+                    <span>+{seeMore}</span>
+                </button>
+            )
+        }
+
+        let fileLabel
+        if (previewFile && previewFile?.type !== 'image') {
```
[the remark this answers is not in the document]
bruno2kd: I'll do `previewFile && previewFile.type`, otherwise ts complains

src/ebay-file-preview-card/file-preview-card.tsx
```diff
@@ -0,0 +1,191 @@
+import React, { FC, useMemo } from 'react'
+import cx from 'classnames'
+import { EbayProgressSpinner } from '../ebay-progress-spinner'
+import { EbayIconButton } from '../ebay-icon-button'
+import { EbayIcon } from '../ebay-icon'
+import { EbayMenuButton, EbayMenuButtonItem } from '../ebay-menu-button'
+import { FilePreviewCardInput, EbayMenuSelectEventHandler } from './types'
+
+const EbayFileInput: FC<FilePreviewCardInput> = ({
+    a11yCancelUploadText,
```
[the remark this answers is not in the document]
bruno2kd: I let that pass

src/ebay-file-preview-card/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,279 @@
...
+                status="uploading"
+                a11yCancelUploadText="cancel upload"
+                onCancel={(e) => action('onCancel')(e)}
+                {...args}
+            />
+        )
+    }
+}
+
+export const Image = {
```
HenriqueLimas, line 120: We should follow same structure as other components for story definition (check [combobox](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-combobox/__tests__/index.stories.tsx#L137))

```jsx
// Not the usage of a function instead of { render }
export const Default: StoryObj<typeof EbayFilePreviewCard> = (args) => (

)
```

And there is no need for fragments in case one single element is rendered. make sure to add types for all the stories

src/ebay-file-preview-card/__tests__/index.stories.tsx
```diff
@@ -0,0 +1,279 @@
...
+        <>
+            <EbayFilePreviewCard
+                a11yCancelUploadText="Cancel upload"
+                deleteText="Delete"
+                file={{
+                    name: 'file-name.jpg',
+                    type: 'image',
+                    src: 'https://ir.ebaystatic.com/cr/v/c01/skin/docs/tb-real-square-pic.jpg'
+                }}
+                onCancel={(e) => action('onCancel')(e)}
```
HenriqueLimas, line 131: These are already handle by `...args` no need to add them

src/ebay-file-preview-card/file-preview-card.tsx
```diff
@@ -0,0 +1,234 @@
...
+            <>
+                <EbayIcon name="file24" className="file-preview-card__asset" />
+                {getPreviewCardAction()}
+                {getFileLabel()}
+            </>
+        )
+    }
+    return (
+        <CardEl className="file-preview-card" {...rest}>
+            <div className="file-preview-card__body">{previewCardContent}</div>
```
HenriqueLimas, line 223: These logic is quite tricky to read. I think we can separate into two components instead. At the end we will use something like these bellow (just to show the idea)

```jsx
return (
   ...
   <div class="file-preview-card__body">
      <FileContent  file={file} status={status} isFade={seeMore !== undefined} />
      {seeMore ? (
           <button></button>
      ) : <FileAction file={file} infoText={infoText} status={status} onCancel menuActions onMenuSelect onDelete />
)
```

Then each component can have the logic inside it 

```jsx
export const FileContent = ({}) => {
  if (status === uploading) {
    return ()
  }
  
  switch(file.type) {
    case 'video':
       return <video />
    case 'image':
       return <img />
    default:
       return <EbayIcon />
  }

}
```

Same for the action. I think this is a more "react" way of solving the problem

HenriqueLimas (approved):

PR #477, opened 2025-04-10
expose dialog footer styling

ericbiewener:
Currently, the dialog footer styling is only available via the `footer={...}` prop on the various dialogs. This is because the `EbayDialogFooter` component does _not_ actually include its styling -- it acts only as [a way for `DialogBaseWithState` to _find_ the footer child](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-dialog-base/dialog-base-with-state.tsx#L23), which in turn [gets rendered in a `div` that has the necessary classname styling](https://github.com/eBay/ebayui-core-react/blob/main/src/ebay-dialog-base/components/dialogBase.tsx#L223-L226).

This is problematic because it forces a specific component hierarchy in order to achieve the desired UX. For example, consider a dialog that is a form. This component hierarchy restriction doesn't allow rendering a form's submit button inside the `<form>` element, since the `EbayDialogFooter` component will not be the direct descendent of `DialogBaseWithState`.

This PR adds the wrapping `div` and classname logic to `EbayDialogFooter` so that consumers can get this styling without the component hierarchy restriction.


.eslintrc
```diff
@@ -34,6 +34,8 @@
     "rules": {
         // note you must disable the base rule as it can report incorrect errors
         "no-use-before-define": "off",
+        "no-nested-ternary": "off",
```
ericbiewener, line 37: I love nested ternaries!

src/ebay-dialog-base/components/dialogBase.tsx
```diff
@@ -16,15 +16,13 @@ import * as keyboardTrap from 'makeup-keyboard-trap'
...
 import { DialogCloseEventHandler } from '../types'
 
-
 export type WindowType = 'compact'
-type ClassPrefix = 'fullscreen-dialog' | 'lightbox-dialog' | 'panel-dialog'
-    | 'drawer-dialog' | 'toast-dialog' | 'alert-dialog' | 'confirm-dialog'
-    | 'snackbar-dialog'
-type ButtonPosition = 'top' | 'right' | 'bottom' | 'left' | 'hidden'
 
+type ButtonPosition = 'right' | 'left' | 'hidden'
```
ericbiewener, line 25: I think the elimination of `bottom` here is desirable & does not require a major version bump. It is not used inside this codebase, and it appears that the `buttonPosition` prop is _not_ exposed by any of the wrapping dialog components. Since `EbayDialogBase` not available to consumers, it is not a breaking change.

HenriqueLimas:
By allowing this, it will allow the wrong rendering of the dialog. For now, to support forms you can use the `form` attribute in the `button` element to reference the correct form. The current dialog logic and components is to make sure that dialogs has the correct structure so it doesn't break in the future with skin changes.

